PR #3, opened 2015-02-24
support custom primitive types

chaowlert:
Hi,

Thank you for your library. I'm using your library and ran into issue where some classes cannot map. But it should be just copied reference to target.

In my scenario, the class is `DbGeography` which is standard EF type. Even it is class, but it represents primitive type similar to string. 
And every time mapper maps this class, it will throw exception, because this class has no constructor.

The example usage is `TypeAdapterConfig.GlobalSettings.PrimitiveType.Add(typeof (DbGeography));`

### DIFF
--- a/src/Fpr/TypeAdapterGlobalSettings.cs
+++ b/src/Fpr/TypeAdapterGlobalSettings.cs
@@ -1,4 +1,7 @@
 ﻿
+using System;
+using System.Collections.Generic;
+
 namespace Fpr
 {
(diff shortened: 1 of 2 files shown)


eswann:
Nice, thanks for the pulls, will review and pull.

eswann:
Hey chaowlert, after pulling these in, I'm getting a number of unit test failures.  Were you able to run unit tests successfully?

PR #4, opened 2015-02-24
change hash key to long

chaowlert:
Hi,

Based on your library, you use `int` as key of record of {TSource + TDestination + has destination flag}. 
Even it is very small chance, but it is possible to have key collision when there are a lot of type mappings.

My solution is using `long`, TSource will use high-order number, TDestination will use low-order number, and destination flag will be positive and negative.

Please review

### DIFF
--- a/src/Fpr/Adapters/ClassAdapter.cs
+++ b/src/Fpr/Adapters/ClassAdapter.cs
@@ -19,7 +19,7 @@ namespace Fpr.Adapters
         private static Func<TDestination> _destinationFactory;
         private static AdapterModel<TSource, TDestination> _adapterModel;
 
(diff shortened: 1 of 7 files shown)


eswann:
This pull reqest caused several tests to fail...had to nix it.  Good idea but did you try running the tests?

eswann:
I was able to fix the failures by updating the signatures of the ClassAdapter AdaptInvoker signatures to be long,int instead of int,int.  Please verify the unit tests going forward, but I agree it's a good update.

chaowlert:
Hi, sorry, I just ignore notification without realize you have comments. I'm so sorry about bug from my PR. For all unit tests, I will do it soon.

eswann:
No problem...thanks for the contributions!

PR #5, opened 2015-02-24
support name with underscore + support ComplexType

chaowlert:
Hi,

This problem comes around Queryable mapping. For EF, `Complex` type is just group of property, it should never be null.

Currently, when we map `src.ContactInfo.Address` to `dest.ContactInfo_Address`, it will generate `ContactInfo_Address = src.ContactInfo == null ? null : src.ContactInfo.Address`.
When we run this query exception occurred, because EF doesn't know how to map `src.ContactInfo` with `null`.

My solution is just check for `Complex` type, and this will generate `ContactInfo_Address = src.ContactInfo.Address` without null checking.

As part of this PR, I also add support for name with underscore when flattening mapping.
Standard EF, for `Complex` type and also navigation type, by default, it will generate name with underscore (`ContactInfo_Address` for my case).
So this fix it allow `ContactInfo.Address` flatten to `ContactInfo_Address`.

Please review.

### DIFF
--- a/src/Fpr/Fpr.csproj
+++ b/src/Fpr/Fpr.csproj
@@ -37,6 +37,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
+    <Reference Include="System.ComponentModel.DataAnnotations" />
     <Reference Include="System.Core" />
     <Reference Include="System.Xml.Linq" />
     <Reference Include="System.Data.DataSetExtensions" />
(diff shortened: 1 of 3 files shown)


eswann:
This last one is a problem.  It's critical that FPR be fast, that's its entire purpose of course.  The added code slows it down ~5%.  Please make sure to run the benchmarks before submitting.  A lot of things in the project are set up a certain way for speed optimization, some things that look like sloppy code are there for measured perf gains.  I'll try to figure out what the culprit is here...but I don't have a ton of time to spend on this.  TrimStart is slow...putting this on every property evaluation is going to slow things down.  There has to be a faster way to accomplish this...I'll see what I find.

PR #15, opened 2016-01-20
Static Adapt methods now work as extension methods and fixed docs where 'OfType' should be 'ForType'.  Also shows Enum -> string failure on compile with Explicit on.

eric-swann-q2:
Updated Adapt static methods to be extension methods on the source so the less terse notation can be used: var dto = product.Adapt<ProductDTO>(config);

Added a few tests for this and also fixed a problem in the documentation: 'OfType' should have been 'ForType' when using instance configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,10 @@ or just
 
     var destObject = TypeAdapter.Adapt<TDestination>(sourceObject);
 
+or using extension methods
+
+	var destObject = sourceObject.Adapt<TDestination>
+
 #####Mapping to an existing object <a name="MappingToTarget"></a>
 You make the object, Mapster maps to the object.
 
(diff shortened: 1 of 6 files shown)


src/Mapster/TypeAdapterConfig.cs
```diff
@@ -312,6 +312,9 @@ internal TypeAdapterSettings GetMergedSettings(Type sourceType, Type destination
         {
             if (this.RequireExplicitMapping && mapType != MapType.InlineMap && !MapContext.HasContext)
             {
+				if(sourceType.)
```
chaowlert: I guess you did not intend to add this line.

README.md
```diff
@@ -89,6 +89,10 @@ or just
 
     var destObject = TypeAdapter.Adapt<TDestination>(sourceObject);
 
+or using extension methods
+
+	var destObject = sourceObject.Adapt<TDestination>
```
chaowlert: This is cool idea.

PR #18, opened 2016-01-21
Explicit mapping

chaowlert:
Hi eric, please help review this PR. This PR included your change on extension method. By merging this PR will accept your PR.

I don't fully understand how RequireExplicitMapping works.
From what I understand after your posted issues is RequireExplicitMapping will only apply to Class (not List and Primitive).

Here is scenario.

```
class Parent {
    public string Id { get; set; }
    public List<Child> Items { get; set; }
}
class Child {
    public List<int> Blah { get; set; }
}

var result = TypeAdapter.Adapt<Parent[], List<Parent>>(src);
```

if we have only `TypeAdaptConfig<Parent, Parent>`, we will throw explicit alert on (Child, Child).

But if we have both `TypeAdaptConfig<Parent, Parent>` and `TypeAdaptConfig<Child, Child>`, mapping will be succeeded (we will not validate list and primitives).

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,10 @@ or just
 
     var destObject = TypeAdapter.Adapt<TDestination>(sourceObject);
 
+or using extension methods
+
+	var destObject = sourceObject.Adapt<TDestination>
+
 #####Mapping to an existing object <a name="MappingToTarget"></a>
 You make the object, Mapster maps to the object.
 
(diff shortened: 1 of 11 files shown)


chaowlert:
PS. Could you help configure AppVayor to build only Mapster.sln?

eric-swann-q2:
I think the behavior you have described here is as designed...I believe this is as it should be in this case.

eric-swann-q2:
I'll take a look at appveyor

eswann:
AppVeyor is fixed...I"m going to merge this pull request and do some more testing on it.

Issue #17, opened 2016-01-20
With explicit mapping turned on...map from IList<TSrc> to List<TDest> fails even though underlying types are mapped.


eric-swann-q2:
Looks like with explicit mapping enabled...a call to map IList<TSrc> to List<TDest> where TSrc and TDest are mapped is throwing.

eric-swann-q2:
This is fixed with latest commits.

eswann:
@chaowlert Thanks a lot for tracking down these issues...very excited that you took over this project and took it to the next level.  I'm excited about using this product again!  Now using it on my new project.

PR #21, opened 2016-01-21
Test demonstrating explicit validation issue between different enumerables with mapped generic types

eric-swann-q2:
Provides a test that demonstrates the problem with validation between different enumerables with the mapped generic types: Issue #17


eswann:
Going to close this one and open a new one after merging latest.

PR #22, opened 2016-01-21
fix code smell, RequireExplicitMapping should belong to ClassAdapter

chaowlert:


eric-swann-q2:
Yeah, I like that change, wasn't happy with the out parameter...looks cleaner now.

PR #25, opened 2016-01-21
Fix Compile break on projection

chaowlert:
Merge this one will also merge #23.

Problem is there is separate logic for projection to do casting only, it will not try to convert type with external method. When we compile, projection is also complied, and exception occurred because we cannot cast enum to string directly.

Original decision which I created separate logic for projection is because if we convert primitive using external method. Sql will not be able to generated.

This PR will make projection and adapt using the same logic. Side effect is people who use projection will get runtime error rather than error on compile method. I use projection a lot. I think it is ok in this case. People who use projection should know that there is limitation in generating sql.


eric-swann-q2:
Cool @chaowlert, thanks for the explanation. Totally understand why you would want to separate projection logic. I agree that this is OK...it would be good to find a way in the future to compile projections separately with slightly different rules.  Thanks for your effort on this.  Do you feel OK with merging/building publishing of this?

PR #26, opened 2016-01-22
Move MapContext usage to inside PreserveReference block

chaowlert:
There is one potential code smell which I don't like. It is the using of static MapContext to share data across mapping process by using ThreadStatic. Currently, only PreserverReference feature uses it.

For this PR, I moved MapContext usage into code block that using PreserveReference. And try to make usage of MapContext safer by implement something similar to Transaction & TransactionScope.

Please help reviewing code in case there is anything concern.


eswann:
Looks good I like the concept and the change...looks like tests are passing...I'm going to merge it.

PR #32, opened 2016-01-25
Add easier support for 3rd party service locators

chaowlert:
Other applications might depend on MEF or other libraries to create catalog of plugins. This PR will make Mapster more friendly to register IRegister.

```
[Export(typeof(IRegister))]
class MyRegister: IRegister
{

}

var catalog = new ApplicationCatalog();
var container = new CompositionContainer(catalog);

var registers = container.GetExports<IRegister>();
TypeAdaptConfig.GlobalSettings.Apply(registers);
```


eswann:
Good idea thanks!

eswann:
@chaowlert can you push a new version out to Nuget?  I would like to pull in these changes but not sure how you are building/publishing.  If you can give me some info on that, would be appreciated....

chaowlert:
@eswann I build using Mapster.xproj (.Net Core proj). Would you like to try? 

I just bump version in project.json and compile in Release mode in Visual Studio. Output nuget will be in artifacts folder.

If you build using CI, you may find command line here http://docs.asp.net/en/latest/dnx/projects.html

Feel free to ask. I can publish if you cannot find the time.

eswann:
Ah ok, previously I was using the psake scripts in the deploy folder which builds and publishes to Nuget...but noticed this wasn't properly setting up the nuget package.

eswann:
I'm going to bump to 2.1.0 per semver b/c we've added some new functionality that's backwards compatible.

eswann:
@chaowlert I'll try to update the package/publish bat files to use dnu...but might wait until RC2 since the CLI is going to be reorganized/finalized at that point.

PR #45, opened 2016-02-16
added dictionary mapping sample

centur:


chaowlert:
Sorry, I will close this one because I would like to support dictionary mapping properly.

centur:
up to you, I was thinking more about providing some working solution before you can fix the issue with the dictionaries in a supported way. From my experience using Mapster - it's good. But lack of working practical examples for mapping complex nested objects like a model with a collection of dictionary entries each of them has it's own nested array or list - is a showstopper. I have to spend time running unit tests for mapping playing with configuration forth and back just to figure out what's wrong in my mappings.

chaowlert:
Good point. I will merge then. Please also help me review PR #46

PR #46, opened 2016-02-16
Add support for record type

### LINKED ISSUE #28: If only have a single constructor with parameters...map to parameters of constructor
I'm currently building a system that is very DDD oriented...as such we only want to initialize objects (even value objects) that are in the domain model (and events) via constructors.  No manual assignment to properties (not even private assignment!)  We don't want to initialize via mapping to properties...period.  Always way to use the constructor  Would be cool if:
1) If object has a single (default) constructor with parameters, attempt to perform standard name/object mapping to the constructor parameters when creating the object.
2) Less important: If there are multiple constructors, possibly allow a specific constructor to be chosen (attribute?) and map to that constructor (Automatically, not using ConstructWith...)

Have too many other things on my plate to implement now...but want to add a placeholder here.

### LINKED ISSUE #44: Still has an issue with mapping dictionaries
Mapster can't map `Dictionary<string,Foo>` to `Dictionary<string,Bar>` although there is a valid mapping between Foo and Bar

This is because Dictionary implements IEnumerable<KeyValuePair> which is immutable.

Throws:
```
Cannot convert immutable type, please consider using 'MapWith' method to create mapping: 
TSource: System.Collections.Generic.KeyValuePair`2[System.String,Foo] 
TDestination: System.Collections.Generic.KeyValuePair`2[System.String,Bar]
```

Simple configuration like this
```
TypeAdapterConfig<Dictionary<string, Foo>, Dictionary<string, Bar>>
  .NewConfig()
  .MapWith(x => x.ToDictionary(k => k.Key, v => v.Value.Adapt<Bar>()))
```
works well and fixes the problem

Is this possible to add such behaviour as default\global for `Dictionary` and `Dictionary<TK,TV>` ? 

As it's a bit inconvenient to explicitly configure all possible Dictionary combinations..

chaowlert:
Fix #44 
Fix #28 

This PR is to support Record Type.

Currently support Record Type with
- No setter
- Have exactly 1 non-empty constructor
- All constructor parameters match with all properties


chaowlert:
@centur 
@eswann 
Could you please help testing this PR for Dictionary & Record Type mapping?

centur:
yes, will do tonight.

centur:
I believe there are no NuGet packages I can use for testing, like some beta feed, right?
What's the best way to help you with this? Can I just build it from source with PR (from your branch) and then run some sample mapping tests to verify it's working ?

chaowlert:
I published pre-release version here https://www.nuget.org/packages/Mapster/2.2.0-beta.
There is changing in core api, I would like to test more than just unit test.

Thank you for help testing.

PR #51, opened 2016-03-03
Add IgnoreCaseSensitiveSetting

BuzzDee3000:
I have implemented the feature to ignore the CaseSensitiv naming of class members.
https://github.com/eswann/Mapster/issues/49

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,6 +369,17 @@ In Mapster 2.0, you can even map when source and destination property types are 
         .Map(dest => dest.Gender,      //Genders.Male or Genders.Female
              src => src.GenderString); //"Male" or "Female"
 
+Mapster can also match properties with a different case sensitivity for a special configuration or on global level. 
+
+    TypeAdapterConfig<TSource, TDestination>
+        .NewConfig()
+        .Settings
+        .IgnoreCaseSensitiveNames = true;
+
+    TypeAdapterConfig
+        .GlobalSettings
+        .IgnoreCaseSensitiveNames = true;
+
 #####Merge object <a name="Merge"></a>
 By default, Mapster will map all properties, even source properties containing null values.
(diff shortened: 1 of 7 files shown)


eswann:
Looks like a good addition....thanks for adding tests...can you add an update to the readme?  Also i can see this as a setting that somebody would like to set globally as well.  That can come later if need be.

BuzzDee3000:
Hi @eswann 
Yes your right with the global setting. I have added this and I also added a new paragraph to the readme.
I was not sure for the correct location in the readme but I hope its fine for you.

chaowlert:
Hi @BuzzDee3000,

Thank you very much for contribution!

I'm also interested in this feature, but after I studied, we should support more than just ignore case. My previous project also had case which we need to map snake case (ie. `first_name` map to `FirstName`).

After I reviewed AutoMapper, what AutoMapper implemented is perfect. AutoMapper provides a set of mapping (pascal case, camel case, prefix, suffix). Moreover, users can implement their own custom name mapping. (https://github.com/AutoMapper/AutoMapper/wiki/Conventions)

This PR is great job. But I just think if we add custom mapping similar to AutoMapper, how setting will look like?

This weekend I will try to work on Mapster, and I will help to extend this feature.

@eswann do you have any idea?

BuzzDee3000:
A simple solution could be to give the user the possibility to register some string compare expressions. So the user has fully control and mapster could provide some common cases.
Something like .AddNameComparson((s,t)=>s.equals(t))
I'm not at home and just have my phone with poor internet so I could not try..

eswann:
@chaowlert, @BuzzDee3000  - I like the idea of string compare expressions with some OOTB solutions.  It would be nice to be able to support all manner of comparisons including snake_case and kebab-case.

chaowlert:
I will close this issue as I created similar functionality in #57.

PR #53, opened 2016-03-05
cloning config

### LINKED ISSUE #41: Consider making TypeAdapterConfig internal state immutable
I need a way to clone the global typeadapterconfig for an instance config and make it efficient.

Here is the scenario:  I have all of my global configuration set for my app.  I want to inject an IAdapter per scope that inherits from the global config but adds an adapter rule just for that adapter instance because it's dependent on an item that's injected.  If the internal state of the TypeAdapterConfig was comprised of immutable collections, it would make clone operations very fast and updates to the resultant config would not impact the global config.
@chaowlert if you have any ideas on this, would be appreciated.

chaowlert:
fix #41 

This PR includes
1. Clone functionalities (`config.Clone()` @eswann pls help checking whether this is what you need or not)
2. Prevent from changing setting after calling to Adapt method
3. Fix public field mapping
4. There are some renaming for internal classes.


eswann:
Hi @chaowlert, thanks for adding this.  I came up with an alternate solution for my issue and I was worried about the performance of cloning a config on the fly. Even with the current solution I believe it would result in significant performance impacts that would negate the value in my case (although I haven't tested it).  
In my case, that's what I was potentially looking at: cloning a config to add in a simple runtime mapping, so this would be doing a ton of clones (per request essentially).  The reason for suggesting immutable collections is that shallow clone functionality for these is extremely fast (as it's just reusing the immutable you are cloning from)...but understood that could result in side effects if you don't understand the implications.  In my case I believe it would have worked as I wanted a shallow clone of the configuration and then to just add a new mapping on the fly (not adjust an existing mapping).  Sounds like the last pull request you added would satisfy this although I haven't looked closely enough to understand the perf implications.

PR #55, opened 2016-03-05
Passing Runtime value

### LINKED ISSUE #48: ForType Settings
Hello. I have some missunderstanding. I have example code

        public class SourceClass
        {
            public string A { get; set; }
            public string B { get; set; }
            public string C { get; set; }
        }
        public class DestClass
        {
            public string A { get; set; }
            public string B { get; set; }
            public string C { get; set; }
        }
        private void MappingTest()
        {
            var sourceItem = new SourceClass()
            {
                A = "A source",
                B = "B source",
                C = "C source",
            };
            
            TypeAdapterConfig<SourceClass, DestClass>.NewConfig().Map(dest => dest.A,
              src => string.Format("{0} {1}", src.A, "Add something To A"));

            // This mapping gives:
            // firstResult.A =  "A source Add something To A"
            // firstResult.B = "B"
            // firstResult.C = "C"
            var firstResult = TypeAdapter.Adapt<DestClass>(sourceItem);

            // This config should extend previous config
            TypeAdapterConfig<SourceClass, DestClass>.ForType().Map(dest => dest.C,
                src => string.Format("{0} {1}", src.C, "Add something To C"));

            // This mapping also gives:
            // secondResult.A = "A source Add something To A"
            // secondResult.B = "B"
            // secondResult.C = "C"
            // But It should give
            // secondResult.A = "A source Add something To A"
            // secondResult.B = "B"
            // secondResult.C = "C Add something To C"
            var secondResult = TypeAdapter.Adapt<DestClass>(sourceItem);

        }

As you see, second `ForType` config is not applied. I resolved this by calling on second (`ForType`) config `Compile()` method. Is it correct?

chaowlert:
Fix #48 

(Please merge #53 first before reviewing this PR.)

This PR includes
1. Passing run-time values to config
2. New builder pattern which allow us to pass run-time config to Adapt method.

**Example**

On configuration, we can receive run-time value by `MapContext.Current.Parameters`.

```
TypeAdapterConfig<Poco, Dto>.NewConfig()
                            .Map(dest => dest.CreatedBy,
                                 src => MapContext.Current.Parameters["user"]);
```

To pass run-time value.

```
var dto = poco.BuildAdapter()
              .AddParameters("user", this.User.Identity.Name)
              .AdaptToType<SimpleDto>();
```

### DIFF
--- a/src/Mapster.Tests/Mapster.Tests.csproj
+++ b/src/Mapster.Tests/Mapster.Tests.csproj
@@ -71,6 +71,7 @@
     <Compile Include="WhenMappingRecordTypes.cs" />
     <Compile Include="WhenMappingIgnoreNullValues.cs" />
     <Compile Include="WhenMappingWithExtensionMethods.cs" />
+    <Compile Include="WhenPassingRuntimeValue.cs" />
     <Compile Include="WhenPreserveReferences.cs" />
     <Compile Include="WhenAddingCustomMappings.cs" />
     <Compile Include="WhenHandlingUnmappedMembers.cs" />
--- a/src/Mapster.Tests/WhenPassingRuntimeValue.cs
+++ b/src/Mapster.Tests/WhenPassingRuntimeValue.cs
@@ -1,0 +1,44 @@
+﻿using System;
+using NUnit.Framework;
+using Should;
+
+namespace Mapster.Tests
+{
+    public class WhenPassingRuntimeValue
+    {
+        [Test]
+        public void Passing_Runtime_Value()
+        {
+            TypeAdapterConfig<SimplePoco, SimpleDto>.NewConfig()
+                .Map(dest => dest.CreatedBy,
+                    src => MapContext.Current.Parameters["user"]);
+
+            var poco = new SimplePoco
+            {
+                Id = Guid.NewGuid(),
+                Name = "test",
+            };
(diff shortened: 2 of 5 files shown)


eswann:
@chaowlert, Thanks for this work...looks good, I'm a fan of the renames and the MapContext functionality.  Have you evaluated the impact to benchmarks?  Is there a significant penalty for including a parameter in the mapping since there are a couple of checks?  Dict null should be negligible but was there a significant impact when there is a param dict lookup?

chaowlert:
@eswann I just ran benchmark. Performance is still the same.

When we use MapContext, there will be performance effect for sure. MapContext is ThreadStatic, accessing will be a bit slower. Plus we need to access value by Dictionary of object, which there are cost for lookup and casting. But if we don't use MapContext, there will be no effect. Anyway, performance impact of using MapContext should be the matter of milliseconds, there should be no concern.

In fact, this feature shares common functionality with preserve reference feature. When I did preserve reference feature, I used to pass parameters as a function argument. But it turned out that, we need to re-engineer a lot which is very high risk. Plus there is performance impact to all other cases by 10-15%, because we need more check for extra arguments. So I decided to sacrifice some good practices for lesser used functionalities, to mitigate risk of re-engineer and impact on performance, by introducing MapContext.

PR #57, opened 2016-03-12
Flexible name matching

### LINKED ISSUE #49: Ignore Case-sensitivity on Propertynames
It would be nice if mapster can ignore differend Upper/Lowercase naming. I am mapping from a Webservice where lots of things are in uppercase (Id vs ID).

I could not found anything anythin in the settings to ignore the case for the matching names.

chaowlert:
Fix #49 

Pls merge #56 before reviewing this issue.

This PR allows match property names with different convention. Usage is

```
TypeAdapterConfig.GlobalSettings.Default.NameMatchingStrategy(NameMatchingStrategy.Flexible);
```

or

```
TypeAdapterConfig<Foo, Bar>.NewConfig().NameMatchingStrategy(NameMatchingStrategy.Flexible);
```

By defining as flexible, it will allow to match between `PascalCase`, `camelCase`, `lower_case`, and `UPPER_CASE`. This feature is applied to both property mapping & class flattening. And users can defined their own convention (ie, prefix, suffix, replace) by create new NameMatchingStrategy.

### DIFF
--- a/src/Mapster.Tests/Mapster.Tests.csproj
+++ b/src/Mapster.Tests/Mapster.Tests.csproj
@@ -72,6 +72,7 @@
     <Compile Include="WhenMappingRecordTypes.cs" />
     <Compile Include="WhenMappingIgnoreNullValues.cs" />
     <Compile Include="WhenMappingWithExtensionMethods.cs" />
+    <Compile Include="WhenMappingWithFlexibleName.cs" />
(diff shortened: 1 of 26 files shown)


chaowlert:
@BuzzDee3000 Could you kindly test this feature whether it is what you want or not?

BuzzDee3000:
@chaowlert I will check this asap.

PR #72, opened 2016-07-14
Fix ArgumentException when compiling a derived class without members.

Duracell1989:


chaowlert:
Thank you for contribution :+1:

Issue #77, opened 2016-08-04
Condition in custom property mapping is not working


satano:
Hi

Condition in custom property mapping is not working - or at least it is not working as expected. The documentation says:

> If the condition is not met, the mapping is skipped altogether.

So as I understand it, if the condition is not met, the value in target property should not be changed. But the behavior is this:
- If the condition is met, mapping is performed as usual.
- If the condition is not met, **the mapping is not skipped, but default value for target property type is used.** This effectively nulls reference types, zeroes numbers etc.

# Example
```csharp
public enum TestEnum
{
    DoNotMap = 0,
    Value1 = 1,
    Value2 = 2,
}

public class TestClass
{
    public TestEnum EnumVal { get; set; }
    public string StringVal { get; set; }
}

public static void TestMapster()
{
    TypeAdapterConfig<TestClass, TestClass>.NewConfig()
        .Map(dest => dest.EnumVal, src => src.EnumVal, src => src.EnumVal != TestEnum.DoNotMap)
        .Map(dest => dest.StringVal, src => src.StringVal, src => src.StringVal != "lorem");

    var testSource = new TestClass() { EnumVal = TestEnum.DoNotMap, StringVal = "lorem" };
    var testDestination = new TestClass() { EnumVal = TestEnum.Value1, StringVal = "ipsum" };

    Console.WriteLine(testDestination.EnumVal); // Value1
    Console.WriteLine(testDestination.StringVal); // ipsum
    TypeAdapter.Adapt(testSource, testDestination);
    Console.WriteLine(testDestination.EnumVal); // DoNotMap (0) - but should be Value1
    Console.WriteLine(testDestination.StringVal?? "* NULL *"); // null - but should be "ipsum"
    Console.ReadLine();
}
```
Stano

satano:
Hi again

I just looked at the source code and according to tests, this behavior is correct. So I have two _issues_ with this:

- I think, the documentation should be changed. For me the sentence **_the mapping is skipped altogether_** really means, that **_nothing is going to be done_**. And not that _destination is going to be nulled out_.
- My personal opinion is, that it really would be helpful to have some mechanism to skip mapping of a property, based on a condition. And I recommend that in this conditional function is available not only source object, but also the destination object. So the condition can be based even on destination values.

Stano

chaowlert:
1. document has been updated, thank you for reporting.
[...]

PR #78, opened 2016-08-05
IgnoreIf() - conditional ignoring of properties mapping

satano:
Addresses the issue #77.

We need to skip mapping according to some condition, but `Map` with condition sets target to `null`, when condition is met.

I implemented `IgnoreIf` configuration. It ignores mapping the similar way as standard `Ignore`, **but only when condition is met.** This condition takes as arguments both, source and destination objects.

```csharp
TypeAdapterConfig<TSource, TDestination>
    .NewConfig()
    .IgnoreIf((src, dest) => !string.IsNullOrEmpty(dest.Name), dest => dest.Name);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,14 @@ Mapster will automatically map properties with the same names. You can ignore me
         .NewConfig()
         .Ignore(dest => dest.Id);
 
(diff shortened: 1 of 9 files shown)


chaowlert:
Great work! thank you very much!

PR #81, opened 2016-08-11
Private fields accessing strategy

hpcsc:
Hi,

I just created a private fields/properties accessing strategy (ValueAccessingStrategy.PrivatePropertyOrField), could you help to take a look?

A few things to note:

- I don't include this accessing strategy in default strategy list, since it might be a special case and I think better let users to choose whether to use it or not
- In ClassAdapter.CanMap(), I choose to ignore all types from System namespace since those should have other adapter handle already
- In ClassAdapter.CanInline(), if destination class has any private property, we cannot use inline expression   and have to use block expression (to use Expression.Assign())
- In ClassAdapter.CanInline() and ClassAdapter.GetClassModel(), probably can optimize further by checking whether user specifies ValueAccessingStrategy.PrivatePropertyOrField. But I'm not sure how to do that without massively change other classes at the moment.

Thanks

### DIFF
--- a/src/Mapster/Adapters/ClassAdapter.cs
+++ b/src/Mapster/Adapters/ClassAdapter.cs
@@ -23,11 +23,10 @@ namespace Mapster.Adapters
             if (sourceType == typeof(string) || sourceType == typeof(object))
                 return false;
 
-            if (destinationType.GetTypeInfo().IsEnum || destinationType.Namespace.StartsWith("System"))
+            if (!destinationType.IsPoco())
                 return false;
 
-            return (destinationType.GetFieldsAndProperties(allowNoSetter: false).Any() ||
-                    destinationType.GetFieldsAndProperties(allowNoSetter: false, isNonPublic: true).Any());
+            return true;
         }
 
         protected override bool CanInline(Expression source, Expression destination, CompileArgument arg)
@@ -37,8 +36,6 @@ namespace Mapster.Adapters
             if (arg.MapType != MapType.Projection &&
                 arg.Settings.IgnoreNullValues == true)
                 return false;
-            if (arg.DestinationType.GetFieldsAndProperties(isNonPublic: true).Any())
-                return false;            
             return true;
         }
 
@@ -128,7 +125,6 @@ namespace Mapster.Adapters
             return new ClassModel
             {
                 Members = destinationType.GetFieldsAndProperties(allowNoSetter: false)
-                    .Concat(destinationType.GetFieldsAndProperties(allowNoSetter: false, isNonPublic: true))
             };
         }
     }
--- a/src/Mapster/Adapters/ClassAdapter.cs
+++ b/src/Mapster/Adapters/ClassAdapter.cs
@@ -26,8 +26,8 @@ namespace Mapster.Adapters
             if (destinationType.GetTypeInfo().IsEnum || destinationType.Namespace.StartsWith("System"))
                 return false;
 
-            return (destinationType.GetPublicFieldsAndProperties(allowNoSetter: false).Any() ||
-                    destinationType.GetPrivateFieldsAndProperties(allowNoSetter: false).Any());
+            return (destinationType.GetFieldsAndProperties(allowNoSetter: false).Any() ||
+                    destinationType.GetFieldsAndProperties(allowNoSetter: false, isNonPublic: true).Any());
         }
 
         protected override bool CanInline(Expression source, Expression destination, CompileArgument arg)
@@ -37,7 +37,7 @@ namespace Mapster.Adapters
             if (arg.MapType != MapType.Projection &&
                 arg.Settings.IgnoreNullValues == true)
                 return false;
-            if (arg.DestinationType.GetPrivateFieldsAndProperties().Any())
+            if (arg.DestinationType.GetFieldsAndProperties(isNonPublic: true).Any())
                 return false;            
             return true;
         }
@@ -127,8 +127,8 @@ namespace Mapster.Adapters
         {
             return new ClassModel
             {
-                Members = destinationType.GetPublicFieldsAndProperties(allowNoSetter: false)
-                    .Concat(destinationType.GetPrivateFieldsAndProperties(allowNoSetter: false))
+                Members = destinationType.GetFieldsAndProperties(allowNoSetter: false)
+                    .Concat(destinationType.GetFieldsAndProperties(allowNoSetter: false, isNonPublic: true))
             };
         }
     }
--- a/src/Mapster/Adapters/ClassAdapter.cs
+++ b/src/Mapster/Adapters/ClassAdapter.cs
@@ -23,10 +23,11 @@ namespace Mapster.Adapters
             if (sourceType == typeof(string) || sourceType == typeof(object))
                 return false;
 
-            if (!destinationType.IsPoco())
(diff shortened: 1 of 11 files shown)


chaowlert:
Hi @hpcsc,

Thank you very much for pull request :+1:

src/Mapster/Utils/ReflectionUtils.cs
```diff
@@ -37,6 +38,24 @@ public static IEnumerable<IMemberModel> GetPublicFieldsAndProperties(this Type t
...
+            var properties = type.GetProperties(BindingFlags.Instance | BindingFlags.NonPublic)
+                .Where(x => allowNoSetter || x.CanWrite)
+                .Select(CreateModel);
+
+            var fields = type.GetFields(BindingFlags.Instance | BindingFlags.NonPublic)
+                .Where(x => (allowNoSetter || !x.IsInitOnly) && !IsBackingField(x))
+                .Select(CreateModel);
+
+            return properties.Concat(fields);
+        }
```
chaowlert: I think we should combined `GetPrivateFieldsAndProperties` and `GetPublicFieldsAndProperties` by sending another parameter `isNonPublic` in order to unify logic.

src/Mapster/Utils/ReflectionUtils.cs
```diff
@@ -37,6 +38,24 @@ public static IEnumerable<IMemberModel> GetPublicFieldsAndProperties(this Type t
...
+
+        private static bool IsBackingField(FieldInfo field)
+        {
+            return field.CustomAttributes.Any(a => a.AttributeType == typeof(CompilerGeneratedAttribute));
+        }
```
chaowlert: I think `IsBackingField` is not needed. It might be too much specific to use case. We can use `IgnoreAttribute` to ignore `CompilerGeneratedAttribute`.

src/Mapster/Adapters/ClassAdapter.cs
```diff
@@ -23,10 +23,11 @@ protected override bool CanMap(Type sourceType, Type destinationType, MapType ma
             if (sourceType == typeof(string) || sourceType == typeof(object))
                 return false;
 
-            if (!destinationType.IsPoco())
+            if (destinationType.GetTypeInfo().IsEnum || destinationType.Namespace.StartsWith("System"))
                 return false;
 
-            return true;
+            return (destinationType.GetPublicFieldsAndProperties(allowNoSetter: false).Any() ||
+                    destinationType.GetPrivateFieldsAndProperties(allowNoSetter: false).Any());
```
chaowlert: Please add new class `ClassWithPrivateMemberAdapter` and override logic there. This will break change current behavior.

chaowlert:
I think we could add `TypeAdapterConfig.GlobalSettings.EnablePrivateMember()` and this could be its content.

```
var rule = new ClassWithPrivateMemberAdapter().CreateRule();
rule.Settings.ValueAccessingStrategies.Add(ValueAccessingStrategies.PrivatePropertyOrField);
this.Rules.Add(rule);
```

chaowlert:
Thank you very much @hpcsc, this is great work :)

src/Mapster/ValueAccessingStrategy.cs
```diff
@@ -11,7 +11,8 @@ namespace Mapster
     public static class ValueAccessingStrategy
     {
         public static readonly Func<Expression, IMemberModel, CompileArgument, Expression> CustomResolver = CustomResolverFn;
-        public static readonly Func<Expression, IMemberModel, CompileArgument, Expression> PropertyOrField = PropertyOrFieldFn;
+        public static readonly Func<Expression, IMemberModel, CompileArgument, Expression> PropertyOrField = PropertyOrFieldFn(false);
+        public static readonly Func<Expression, IMemberModel, CompileArgument, Expression> PrivatePropertyOrField = PropertyOrFieldFn(true);
```
chaowlert: This should rename to `NonPublicPropertyOrField` to be exact.

src/Mapster/Utils/ReflectionUtils.cs
hpcsc: could you elaborate a bit more on `IgnoreAttribute`? I'm not aware of any .NET attribute with the name `IgnoreAttribute`, unless it's from NUnit. Or do you mean we create a custom attribute to ignore some private fields? The reason I don't want to include backing field is to avoid duplication, since we already get normal properties which use those backing fields

src/Mapster/Adapters/ClassAdapter.cs
hpcsc: sure, I'll create a new class inheriting from ClassAdapter. I thought about it before, but since I'm not too sure about usage of the score in each adapter class (as I understand, it's to determine precedence of which adapter to check first, but then the magic number for each score might have some story behind it, so I didn't want break something). Which score should I give to this new class? 
Also, is there any specific behavior that you are afraid that might break existing logic if we put the code like above? All the unit tests are passing, so at least not something that the library is covering I guess?

src/Mapster/Utils/ReflectionUtils.cs
chaowlert: Ah sorry, I mean we can change from hard code `CompilerGeneratedAttribute` to config `TypeAdapterConfig.GlobalSettings.Default.IgnoreAttribute(typeof(CompilerGeneratedAttribute))`.

src/Mapster/Utils/ReflectionUtils.cs
hpcsc: I already merged these 2 methods in the recent commits

src/Mapster/Utils/ReflectionUtils.cs
hpcsc: I already merged the 2 mentioned methods, but I'm not entirely sure what's the best way to access global settings from ReflectionUtils, should I just access it directly? It might create unwanted dependency there.

src/Mapster/Adapters/ClassAdapter.cs
hpcsc: I added ClassWithNonPublicMemberAdapter and reverted back the ClassAdapter. For now the score for both classes are the same. Please let me know if further changes needed.

src/Mapster/ValueAccessingStrategy.cs
hpcsc: this is done in recent commits

src/Mapster/Utils/ReflectionUtils.cs
chaowlert: I mean, we should remove `IsBackingField` method completely from Mapster. And we can ignore attribute in application.

src/Mapster/Adapters/ClassAdapter.cs
chaowlert: Yes, score is precedence, higher score will apply when it pass `CanMap` method. I think score for `ClassWithNonPublicMemberAdapter` could be any, since there is no adapter that can map private fields.

chaowlert:
Hi @hpcsc,

I did some refactoring, and move `EnableNonPublicMembers` to specific type setting. I created PR at https://github.com/hpcsc/Mapster/pull/1/files, please help review and test.

hpcsc:
@chaowlert thanks a lot, that's indeed cleaner. I already merged your commit

PR #95, opened 2016-12-07
Object to primitive type is always converted using Convert class. Fixes #94

### LINKED ISSUE #94: InvalidCastException from int to long
I would have thought that assigning an int to a long should work. 
However, this example throws an InvalidCastException:

```
        class FormatOptions
        {
            public int CharactersPerInch { get; set; }
            public long LinesPerPage { get; set; }
        }

        [Fact]
        public void ConvertDictionaryOptions() {
            var exportFormatOptions = new Dictionary<string, object>
                    {
                        { "CharactersPerInch", 16 },
                        { "LinesPerPage", 0 }
                    };
            var options = exportFormatOptions.Adapt<FormatOptions>();
        }
```

satano:
Casting to primitive types from object may fail in some cases as described in #94

```csharp
object intValAsObject = 123; // Value is Int32
long longVal = (long)intValAsObject; // Throws InvalidCastException
```

I have changed `ReflectionUtils.BuildUnderlyingTypeConvertExpression` method so it always use `Convert` class (instead of casting) when source type is `object` and destination type is primitive.

### DIFF
--- a/src/Mapster.Tests/Mapster.Tests.csproj
+++ b/src/Mapster.Tests/Mapster.Tests.csproj
@@ -71,6 +71,7 @@
     <Compile Include="Classes\TypeTestClass.cs" />
     <Compile Include="WhenCloningConfig.cs" />
     <Compile Include="WhenCompilingConfig.cs" />
+    <Compile Include="WhenConvertingFromObjects.cs" />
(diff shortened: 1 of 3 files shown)


chaowlert:
Thank you @satano. Great works! Looking up by dictionary is great!

PR #106, opened 2017-03-29
enum to enum support

klesouza:


chaowlert:
Thank you very much I will publish soon

chaowlert:
@klesouza I cannot publish yet. After some tests, there is breaking change. Original behavior when we map enum to enum, it will map by int value. And after this merge, it will map by name. May be we should have new configuration to map by int value or by name.

klesouza:
That's a good idea, I'll try to think about how to do it

PR #113, opened 2017-06-08
Multidimensional arrays are correctly mapped.

### LINKED ISSUE #99: "Argument must be single dimensional array type" Exception when trying to map type with multi-dimensional array property
Very simple example:
```C#
private static void Main()
{
    var a = new Foo { Ints = new int[2, 1] };
    var b = new Bar { Ints = new int[1, 2] };

    //throws exception
    TypeAdapter.Adapt(a, b);
}

internal class Foo
{
    public int[,] Ints { get; set; }
}

internal class Bar
{
    public int[,] Ints { get; set; }
}
```
I agree that multi-dimensional arrays are not the best thing, but still, even a reference copy would be a better behavior. The only workaround is to override the mapping with .MapWith but then again, see the other issue i'm reporting about MapToTargetWith.

satano:
Fixes #99 - "Argument must be single dimensional array type"

Multidimensional arrays are now correctly created. The significant changes are:

- `Expression.NewArrayBounds()` for instantiating array uses overload which takes arguments for every array's rank.
- Expression for copying array's data (`Array.Copy()`) does not use `Expression.ArrayLength()`, but `arrayVar.LongLength` property.

### DIFF
--- a/src/Mapster/Adapters/CollectionAdapter.cs
+++ b/src/Mapster/Adapters/CollectionAdapter.cs
@@ -22,20 +22,15 @@ namespace Mapster.Adapters
 
         private static Expression CreateCountExpression(Expression source, bool allowCountAll)
         {
-            if (source.Type.IsArray)
-                return Expression.ArrayLength(source);
-            else
-            {
-                var countProperty = source.Type.GetProperty("Count");
-                if (countProperty != null)
-                    return Expression.Property(source, countProperty);  //list.Count
-                if (!allowCountAll)
-                    return null;
-                var countMethod = typeof(Enumerable).GetMethods()
-                    .First(m => m.Name == nameof(Enumerable.Count) && m.GetParameters().Length == 1)
-                    .MakeGenericMethod(source.Type.ExtractCollectionType());
-                return Expression.Call(countMethod, source);            //list.Count()
-            }
+            var countProperty = source.Type.GetProperty("Count");
+            if (countProperty != null)
+                return Expression.Property(source, countProperty);  //list.Count
+            if (!allowCountAll)
+                return null;
+            var countMethod = typeof(Enumerable).GetMethods()
+                .First(m => m.Name == nameof(Enumerable.Count) && m.GetParameters().Length == 1)
+                .MakeGenericMethod(source.Type.ExtractCollectionType());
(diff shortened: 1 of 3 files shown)


satano:
I messed something. Testds for collections are not passing. Sorry.

satano:
I don't understand, why AppVeyor is failing. On my computer, everything is OK. All tests are passing, all benchmarks are running.

satano:
One more thing. It would be nice, if source code was correctly formatted (default formatting in Visual Studion). There is a lot of whitespaces in source files. Getting rid of them is easy - just use `Ctrl+K, D` shortcut and code is formatted. For whitespaces at the end of lines, I recommend [Trailing Whitespace Visualizer](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer) extension. It shows whitespaces at the end of lines in source and removes them on save.

Because now, there is much more differences in files the I really made.

chaowlert:
Thank you @satano, you are helpful as always. 

I think AppVeyor has problem with new C#. And for style different, I think it is because of Resharper. I will try to fix style setting to match with Visual Studio and definitely install the trailing white-space extension.

src/Mapster/Adapters/CollectionAdapter.cs
```diff
@@ -51,47 +46,62 @@ protected override bool CanInline(Expression source, Expression destination, Com
...
         {
-            var destinationElementType = arg.DestinationType.ExtractCollectionType();
-            if (arg.DestinationType.IsArray)
-                return Expression.NewArrayBounds(
-                    destinationElementType, 
-                    CreateCountExpression(source, true));   //new TDestinationElement[count]
+            if (arg.SourceType.IsArray && arg.DestinationType.IsArray)
+            {
+                return CreateArrayInstantiationExpression(source, arg);
+            }
```
chaowlert: Will this break change? I haven't test yet but seemed like we cannot map non-array to array any more.

satano: I will investigate it.

chaowlert: Thank you @satano.

For `List<List<int>> to int[,]`, I think this feature provides too little value but required fair amount of effort. Unless you need this, you not need to work on it.

PR #124, opened 2017-09-27
Map to IEnumerable class is done using ClassAdapter instead of PrimitiveAdapter

### LINKED ISSUE #123: IEnumerable - Mapster.CompileException
If I try to adapt one class to another, which implements IEnumerable interface, I get exception:

Mapster.CompileException
InvalidOperationException: Cannot convert immutable type, please consider using 'MapWith' method to create mapping

Simple example:

```
using Mapster;
using System;
using System.Collections;

namespace MapsterIEnumerableBug
{
    class Program
    {
        static void Main(string[] args)
        {

            var classA = new ClassA();

            var classB = classA.Adapt<ClassB>(); //this throws an exception
        }
    }

    public class ClassA
    {
        public int ID { get; set; }
    }

    public class ClassB : IEnumerable
    {

        public int ID { get; set; }

        public IEnumerator GetEnumerator()
        {
            throw new NotImplementedException();
        }
    }
}
```

satano:
This PR fixes #123. The problem is, that `IEnumerable` classes are not recognized as POCO and thus they are adapted using `PrimitiveAdapter` instead of `ClassAdapter`. I think that removing this condition is safe. If both classes are `IEnumerable`, `CollectionAdapter` will be used - it has higher priority than `ClassAdapter`. And basically, class which implements `IEnumerable` still is just a class and can be mapped as such, and not as a primitive type.

I am not sure what exactly is record type, but I made the same change for `RecordTypeAdapter`.


chaowlert:
thank you very much @satano. you are brilliant.

chaowlert:
cant merge yet AppVeyor hasnt set properly. ill do later.

satano:
The problem is using `out var varName` in function calls. AppVeyor is probably configured to use Visual Studio 2015. It must use Visual Studio 2017.

chaowlert:
Hi I have no time to inspect but seemed like unit tests do not pass.

satano:
Well. All tests are passing on my computer and also in AppVeyor connected to "my" Mapster.

If you could add me as a collaborator to Mapster in your AppVeyor, I could check it out.

satano:
Hi again. Apparently I did not have pulled all of changes in your `master`. When I synced your master, even my tests were failing. The problem was in method `ReflectionUtils.IsRecordType`, where in some cases constructor parameter had no name (was `null`) and so there was `NullReferenceException` in `p.Name.ToPascalCase()`.

I created another PR #125, which fixes this as well

chaowlert:
Hi @eswann @eric-swann-q2,

Pls help following

1. Add me (@chaowman) and @satano to AppVeyor to help fixing build pipeline.
2. Could you pls transfer this repo to https://github.com/MapsterMapper organization? You will still have complete control over the repo, but it will allow other contributors to make change the setting.

eswann:
Will do



Sent from my Sprint Samsung Galaxy S8.

-------- Original message --------

eswann:
Ran into an issue.  Looks like I don't have permissions to transfer ownership to MapsterMapper.  I this error: "You don’t have the permission to create repositories on MapsterMapper".  So looks like I need "Create" permission to transfer.  I also looked into AppVeyor.  Based on the documenation linked below, we should just drop the current AppVeyor build and create a new one under the organization account.  It just takes a few minutes to set up a new build in AppVeyor.  That can be done once we get ownership transferred. https://www.appveyor.com/docs/team-setup/

chaowlert:
Hi @eswann, there is pending invitation for you to MapsterMapper org. My guess is you might need to accept invitation first before transfer. I also sent another invitation to your email.

eswann:
Cool, I found the old invite and accepted.  I've transferred the repository.

chaowlert:
Thank you, I will update Nuget

PR #129, opened 2017-11-04
Fixed #117 - Type does not have a default constructor, when map to target instance

### LINKED ISSUE #117: Type does not have a default constructor
I observed a regression when I switched from v2 to v3: v3 requires that types have a default constructor.

It is not explicited on documentation that this use case is not supported:

	
	void Main()
	{
		var foo = new Foo { Value = "test" };
		var bar = new Bar(something: 1);
		foo.Adapt(bar);
	
		// V2: ok
		// V3: not ok
		foo.Adapt<Foo, Bar>(bar);
	}
	
	
	 class Foo
	{
		public string Value { get; set; }
	}
	
	 class Bar
	{
		public string Value { get; set; }
	
		public int Something { get;}
		
		public Bar(int something)
		{
			Something = something;
		}
	}

Burgyn:
This PR fixes #117. I tried investigate, why and when the problem was arose. Genesis of this is complicated. In commite 8c8d95e, condition (if destination is not null) was removed from method `CreateBlockExpressionBody` in `BaseAdapter` and `CreateInstantiationExpression` was called always. Where default constructor was calling. (what is incorrect when mapping to target instance without default constructor).

I tried this condition give to back, but tests for mapping to target instance failed.

So
I resolved it with condition for default constructor. If type doesn't have default constructor, destination expression is returning from `CreateInstantiationExpression` method.

### DIFF
--- a/src/Mapster.Tests/WhenUsingNonDefaultConstructor.cs
+++ b/src/Mapster.Tests/WhenUsingNonDefaultConstructor.cs
@@ -59,6 +59,18 @@ namespace Mapster.Tests
             dto.Unmapped.ShouldBe("unmapped");
         }
 
+        [TestMethod]
+        public void Map_To_Existing_Destination_Instance_Should_Pass()
+        {
+            var simplePoco = new SimplePoco { Id = Guid.NewGuid(), Name = "TestName" };
+
+            var dto = new SimpleDtoWithoutDefaultConstructor("unmapped");
+            simplePoco.Adapt(dto);
+
+            dto.Id.ShouldBe(simplePoco.Id);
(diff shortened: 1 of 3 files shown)


chaowlert:
Thank you for issue & PR. After merge I will fix a bit. When target object is null and there is no default constructor, we should throw descriptive error.

Burgyn:
Thank you for merge.

I thought about it. I wanted to use it similarly to the abstract class. But when I wrote the test, it actually fell on the type mismatch. I did not like it. I can do this by throwing the exception directly, not the expression.

Can I do that?

PR #217, opened 2020-01-10
Performance tests adapted to use BenchmarkDotNet

SergerGood:


chaowlert:
Superb! Much nicer, thank you!

chaowlert:
image

Please help me on reading results, does this mean CodeGen is faster than AutoMapper by about 20x? And using less memory by 4x?

SergerGood:
> image
> 
> Please help me on reading results, does this mean CodeGen is faster than AutoMapper by about 20x? And using less memory by 4x?

yes :)

for edit reamdme.md you can use the finished results that are in the folder bin\Release\netcoreapp2.0\BenchmarkDotNet.Artifacts\results\*-report-github.md

PR #221, opened 2020-01-18
Support map directly to interface

### LINKED ISSUE #188: Mapping to inherited interface not working
I have the following mapping config defined
 
    config.NewConfig<OrderStatusSaveDto, IOrderStatus>()
        .ConstructUsing(src => null);

(**NB** ConstructUsing(src => null) is because I don't want the mapper to be in control of constructing the interface, it will always be passed an instance)

And then in a controller I receive an `OrderStatusSaveDto` from the post body (which is populated). I then retrieve an `IOrderStatus` entity from a service and call the following to map the properties over

    orderStatus.Adapt(entity);

After this call however, the entity hasn't been populated at all.

Should Mapster be able to map to an interface like this?

### LINKED ISSUE #220: Support mapping directly to interface (without needing the class)
## Feature request

Support mapping directly to some interface, without needing some class implementing that interface. It will create such a class dynamically.

``` csharp
public interface IContract
{
    int Id { get; set; }
    string Value { get; set; }
}

public class ComplexDto
{
    public int Id { get; set; }    
    public string Value { get; set; }
    public string FirstName { get; set; }
    public string LastName { get; set; }
}

var dto = new ComplexDto {
    Id = 1,
    Value = "Test",
    FirstName = "John",
    LastName = "Doe"
};

IContract contract = dto.Adapt<IContract>();
```

`contract` will be an instance of some dynamically generated class implementing the interface `IContract` and its `Id` and `Value` are mapped from `dto`.

## Reasoning

We use service bus in our ecosystem and the bus messages are designed as (simple) interfaces in some base library. We need to convert (map) data to those interfaces. But it is unnecessary to have a class for every interface.

satano:
Closes #220 
Closes #188 

With this changes, Mapster is able to map directly fo interface, without needing the class.

- Mapster will create the class (type) dynamically for every interface it is mapping to. The class is created only once and cached.
- Although the intended use case are simple interfaces with properties only, it supports even interfaces with methods. Calling methods in generated class wil just throw `NotImplementedException`.
- Generics are not supported.
- Indexed properties are not supported.

This (when turned on) will fix the issue #188.

Mapping to interfaces which are not inherited, works out-of-the-box. If interface is inherited, the default behavior stays the same as now - only the properties owned by targe interface are mapped. To map properties from parent interface(s), the setting `UseInterfaceHierarchy` must be enabled.

``` csharp
TypeAdapterConfig.GlobalSettings.Default.UseInterfaceHierarchy(true);
```

The setting can be used as any other setting for specific mapping only.

Personally, I was surprised by this behavior with inherited interfaces. I would expect that all the properties up the hierarchy just work. And so I would prefer default value of `UseInterfaceHierarchy` to be `true`. But this could introduce breaking change for some scenarions.

### DIFF
--- a/src/Mapster.Tests/DynamicTypeGeneratorTests.cs
+++ b/src/Mapster.Tests/DynamicTypeGeneratorTests.cs
@@ -1,0 +1,114 @@
+﻿using Mapster.Utils;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+using System;
+
+namespace Mapster.Tests
+{
+    [TestClass]
+    public class DynamicTypeGeneratorTests
+    {
+
+        private interface INotVisibleInterface
+        {
+            int Id { get; set; }
+            string Name { get; set; }
+        }
+
+        public interface ISimpleInterface
+        {
+            int Id { get; set; }
+            string Name { get; set; }
+        }
+
+        public interface IInheritedInterface : ISimpleInterface
+        {
+            int Value { get; set; }
+        }
+
+        public interface IComplexBaseInterface
+        {
+            int Id { get; set; }
+            void BaseMethod();
+        }
+
+        public interface IComplexInterface : IComplexBaseInterface
+        {
+            string Name { get; set; }
+            int ReadOnlyProp { get; }
+            int WriteOnlyProp { set; }
+            void SimpleMethod();
+            int ComplexMethod(byte b, ref int i, out string s);
+        }
+
+        [TestMethod]
+        public void ThrowExceptionWhenInterfaceIsNotVisible()
+        {
+            void action() => DynamicTypeGenerator.GetTypeForInterface(typeof(INotVisibleInterface));
+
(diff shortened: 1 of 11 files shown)


chaowlert:
Thank you very much! This PR unlocks new possibilities! Let me test a bit and I will release soon.

chaowlert:
Here is my opinion to  `UseInterfaceHierarchy`
1. I cannot see which use case that we will map only first level interface (`UseInterfaceHierarchy` is false).
2. As you said, default behavior, `UseInterfaceHierarchy` should be true

Because of 2, we need to bump version anyway, however, because of 1, I don't think we should introduce more setting.

What is your opinion? Should we remove `UseInterfaceHierarchy`, and make it always true?

satano:
Hi @chaowlert 

I agree that if someone is using an interface, it expects to map all of its properties, not only the first level. So I agree with removing `UseInterfaceHierarchy` setting.

Issue #240, opened 2020-05-11
AdaptMember Attribute stops working when using NameMatchingStrategy


prgood:
When a NameMatchingStrategy is set the AdaptMember Attribute don't work as expected, because the strategy is applied to member names but not to AdaptMember name.

I'll try to do a pull request to fix this. This is my first time :smile:

Here is a test case that fails:

````c#
[TestMethod]
        public void Using_Attributes_With_NameMatchingStrategy()
        {
            TypeAdapterConfig.GlobalSettings.Default.NameMatchingStrategy(NameMatchingStrategy.IgnoreCase);
            var id = Guid.NewGuid();
            var poco = new SimplePoco(id) { Name = "test" };
            var dto = poco.Adapt<SimpleDto>();
            dto.Id.ShouldBe(id);
            dto.Name.ShouldBeNull();
        }

        public class SimplePoco
        {
            public SimplePoco(Guid id) { this.id = id; }

            [AdaptMember("Id")]
            private Guid id { get; }

            [AdaptIgnore]
            public string Name { get; set; }
        }

        public class SimpleDto
        {
            public Guid Id { get; set; }
            public string Name { get; set; }
        }
````

PR #241, opened 2020-05-11
Fix property name match when using AdaptMember Attribute

prgood:
Hi!

This is my first PR. This is supposed to fix: #240 

I had to create a new class to make the test fail. 

If this test runs together with the test on WhenUsingAttribute it wont faild. Maybe because the NameMatchingStrategy setting has no effect after a "Adapt" has ran.

PS: Sorry if my English writing has some mistakes.

### DIFF
--- a/src/Mapster.Tests/WhenUsingAttributeWithNameMatchingStrategy.cs
+++ b/src/Mapster.Tests/WhenUsingAttributeWithNameMatchingStrategy.cs
@@ -1,0 +1,45 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Mapster.Tests
+{
+    [TestClass]
+    public class WhenUsingAttributeWithNameMatchingStrategy
+    {
+        [TestMethod]
+        public void Using_Attributes_With_NameMatchingStrategy()
+        {
+            TypeAdapterConfig.GlobalSettings.Default.NameMatchingStrategy(NameMatchingStrategy.IgnoreCase);
+
+            var id = Guid.NewGuid();
+            var poco = new SimplePoco(id) { Name = "test" };
+            var dto = poco.Adapt<SimpleDto>();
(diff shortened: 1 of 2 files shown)


chaowlert:
Thank you for PR, I will validate soon.

chaowlert:
Cool, I'm quite impress you can capture logic very well, and the fix is clean. 

When we use `IgnoreCase`, how Mapster do comparison is it will lower the case and compare. So in this case, we also need to lower case for name in `AdaptAttribute`.

prgood:
Thanks for the comments and for accepting my PR!

I have some unrelated questions. I will write here because I don't know the right place for it. If necessary I can move it to another place.

I want to improve the CodeGen project so that it can generate code for all mapping and inject this code into Mapster, so that when this feature is enabled all native Adapt, Map e etc use the generated code instead of internal Expression. 

The questions are: Are there some WIP related to it? Is it useful? Would you accept a PR like this?

Thanks

chaowlert:
I'm not sure you mean this or not https://github.com/MapsterMapper/Mapster/wiki/CodeGen.

However I'm not quite like text template. If I have a chance, I would do command line similar to EF migration.

Another thing, SourceGenerator also on radar.

chaowlert:
For the idea about generate mapping, my idea is we could provide interface for example `IMapper<,>`. And let user create empty interface, for example.

```csharp
public interface ICustomerMapper : IMapper<Customer, CustomerDto> { }
```

So user can use their interface to complete their logic upfront.

And on build time, code generator could scan all interfaces implemented `IMapper<,>`. And user will get code that implement `ICustomerMapper`. Mapster could be only tool, not the library for the project.

In fact, what I described above could be done with pure text template, but you need to be a bit expert on text template.

chaowlert:
For your question,

> Are there some WIP related to it?

No, currently my focus is on Azure.

> Is it useful?

If it's useful for you, it will useful for other.

> Would you accept a PR like this?

Why not? However, if the changes introduce a lot of dependencies, I would suggest, we change only core, where it allow to switch implementation. And then, we provide another library to inject new implementation.

prgood:
Great! Thanks for the feedback.

I will follow all your guidelines if I my project really need this type of codegen mapping,

Thanks!

PR #310, opened 2021-02-23
Add test which reproduces an issue with nested async mappings

cmann-andagon:


chaowlert:
Thank you for PR

VladVorobyev:
Hello, and how can I Fix this Issue? It's so powerful option to map like this.

PR #333, opened 2021-05-06
Upgrade packages

SergerGood:


chaowlert:
I'm ok with updating benchmark or testing projects. But I'm not sure we should bump version for production related.

SergerGood:
I rolled back the production versions.

But why not?
For example Mapster.Async 7.2.0 refers to Mapster 7.1.5.
In general, updating package versions is a practice that helps to work with more stable versions.

chaowlert:
Totally agree to update version. I also just bump Mapster to 7.1.5 (latest on that time) for Async module when it released.

However:
1. I have no plan to release the other packages, so let's keep versioning the same as package for now.
2. Bumping version might require testing. It is more effective to test together with new features (or bug fixing). So for now, we have no reason to touch it.

PR #378, opened 2021-10-04
Added support for internal constructors that are accessible via InternalsVisibleToAttribute

jpeters-programmer:
Added support for detecting internal default constructors in the source assembly that are accessible to the destination assembly via the InternalsVisibleToAttribute.  I have not ran the tests in net1_3.  I updated the tests project to target netcoreapp2.1 as otherwise my tests would not run, despite installing the 2.0 SDK.


andrerav:
@jpeters-programmer netcoreapp2.1 is out of support, would you consider changing the framework moniker to net6.0?

andrerav:
@jpeters-programmer The test project has been upgraded to .NET 6.0. Please consider syncing your branch to resolve the current conflict so this PR can be reviewed.

PR #379, opened 2021-10-04
Updated Sample Code in Readme

CoSJay:
The sample code for the queryable example was wrong (typo). (Yes, it's a nothing change, but I know "beginner me" would have been thrown by that.)


andrerav:
Good catch, thank you @CoSJay!

PR #386, opened 2021-11-04
cleaning up

MuhammedJavad:


src/Sample.CodeGen/Models/StudentMapper.g.cs
```diff
@@ -71,17 +70,17 @@ public static Student AdaptTo(this StudentMerge p8, Student p9)
             }
             Student result = p9 ?? new Student();
             
-            if (p8.LastName != null)
+            if (!string.IsNullOrEmpty(p8.LastName))
```
andrerav, line 73: In most cases it is good practice to use string.IsNullOrWhitespace() instead of string.IsNullOrEmpty. I believe this is one of those cases. Will you consider using string.IsNullOrWhitespace()?

PR #409, opened 2022-03-02
[WIP] Add source generators and some other stuff.

Summpot:
I added a source generator to implement the basic models generator (just copy all the properties of the original type).

I would like to use this Pull Request to track the next work.

To Do:

- [x] Change the package reference to Mapster to a project reference.

- [x] Update dependencies.

- [x] Add a basic source generator template.

- [x] Add the Models source generator. 

- [ ] Add the Mappers source generator. 

- [ ] Add the Extensions source generator. 

- [ ] Add custom source generator configuration.

- [ ] Add versioning. (Using GitVersion)

- [ ] Add continuous integration and continuous delivery. (using Github Actions)


andrerav:
Wow, this is an amazing contribution @Summpot! Is there anything you'd like me to help out with? :)

Summpot:
I don't know much about the specific process that Mapster generates the corresponding Model file, and I'm not familiar with the role of configurable options.
 So the progress of this Pull request may not be too fast, can you help me when adding configurable options later, thank you very much.

Summpot:
@andrerav 
I have an issue where the execution of source generators is out of order, which means that one source generator cannot access the generated code of another source generator.
I have two ideas now, either generate the model and mapper in the same source generator, or generate AdaptXX as a member method of the model.
Do you have any suggestions?

Issue #406, opened 2022-02-23
UseDestinationValue is not respected by ClassAdapter:CanMap when implicit mapping


foriequal0:
We use this setting https://github.com/MapsterMapper/Mapster/wiki/Mapping-readonly-prop
When a class only contains read only props that are designated by `UseDestinationValue`, ClassAdapter:CanMap returns false.

also it can't convert to empty class

stagep:
In order to identify issues that are still active, we are closing issues that we believe are either resolved or are dormant. If your issue is still active then please reopen. Thanks.

PR #415, opened 2022-03-24
Fix RequireDestinationMemberSource for immutable UseDestinationValue field

foriequal0:


andrerav:
@foriequal0 Thank you very much, is this a fix for #406?

foriequal0:
No it isn't. This is a fix for an unreported issue.

andrerav:
Thank you, I will push a prerelease package which includes this fix shortly.

PR #431, opened 2022-04-28
Added IOrderedQueryable Interface on MapsterQueryable 

noelmugnier:
I'm currently using Mapster as mapping tool for OData in order to optimize my sql queries with entity framework and avoid exposing the DAL :)

But when I'm using new Mapper().From(IQueryable<T> query).ProjectTo<U>(); I can't OrderBy the result of the projection, it says can't cast MapsterQueryable<U> to IOrderedQueryable<U>.

So I've added the interface and one unit test to ensure it's possible now. 
Regards.


andrerav:

PR #436, opened 2022-05-09
Support ConstructUsing in CollectionAdapter

devbased:
I think the CollectionAdapter should respect ConstructUsing so the user can control how the collection is instantiated.


andrerav:

Issue #430, opened 2022-04-27
Meaning of mapping to an existing object?


zhenlei520:
Meaning of mapping to an existing object?

I used map to existing object, but I found that it doesn't preserve the value of the existing object during the mapping process, it initializes the value of the existing object, and when the existing object is a collection, the value of the existing object is the same as The response data is not the same, but I think their result should be the same

``` C#
[TestMethod]
public void TestMapToExistingObject()
{
    List<UserDto> sources = new List<UserDto>()
    {
        new()
        {
            Name = "Jim",
            Age = 18
        }
    };

    List<UserDto> destination = new List<UserDto>();
    sources.Adapt(destination);
    Assert.IsTrue(destination.Count == sources.Count);//But it is not established at present, I think he is a bug
}

public class UserDto
{
    public string Name { get; set; }

    public int Age { get; set; }
}
```

sgf:
update existing colliection
 i think Mapster need add new api something like ByField() (search some key to update) 
,and insert new when not exists.

stagep:
In order to identify issues that are still active, we are closing issues that we believe are either resolved or are dormant. If your issue is still active then please reopen. Thanks.

PR #437, opened 2022-05-09
Add destination parameter to before and after mapping delegates

devbased:
I have noticed that the result object is different from the destination but I have no way to do anything about it. So I thought it would be useful to pass the destination parameter to (before|after)mapping delegates.
It could help with cases like #430


andrerav:
Thank you @devbased! This is a solid contribution. Do you have time to update the wiki @ https://github.com/MapsterMapper/Mapster/wiki/Before-after-mapping so it's up to date with this functionality?

andrerav:

devbased:
@andrerav Wiki updated

andrerav:
@devbased Excellent, looks great! Thank you very much! :)

Issue #368, opened 2021-08-26
Cannot find assembly path when building on Linux(Debian)


JoaoRuani:
Hello, I have a asp.net core 5 api using Mapster and Mapster.Tool, that's the csproj where mapster is referenced

**Domain.csproj**
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="FluentValidation.AspNetCore" Version="10.3.0" />
    <PackageReference Include="Mapster" Version="7.2.0" />
  </ItemGroup>

  <ItemGroup>
    <Generated Include="**\*.g.cs" />
  </ItemGroup>

  <Target Name="Mapster" AfterTargets="AfterBuild">
    <Exec WorkingDirectory="$(ProjectDir)" Command="dotnet tool restore" />
    <Exec WorkingDirectory="$(ProjectDir)" Command="dotnet mapster model -a &quot;$(TargetDir)$(ProjectName).dll&quot;" />
    <Exec WorkingDirectory="$(ProjectDir)" Command="dotnet mapster extension -a &quot;$(TargetDir)$(ProjectName).dll&quot;" />
    <Exec WorkingDirectory="$(ProjectDir)" Command="dotnet mapster mapper -a &quot;$(TargetDir)$(ProjectName).dll&quot;" />
  </Target>

  <Target Name="CleanGenerated">
    <Delete Files="@(Generated)" />
  </Target>

</Project>

```
I'am using **Mapster.Tool 8.2**.

Inside Domain's project the files are organized like below

Entities Folder: 
- Users
    - User.cs -> with [AdaptTo("[name]CodeGenDto"),GenerateMapper]
    - UserValidator.cs -> FluentValidation validator

Build on windows runs as expected and the dtos are generated. 

When trying to build on debian the following error occurs at the step `dotnet mapster model -a &quot;$(TargetDir)$(ProjectName).dll&quot;"`:

```
Microsoft (R) Build Engine version 16.11.0+0538acc04 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored /home/joao/dpOfficer.digital.backend/dpOfficer.Api.BusinessLogic/dpOfficerApi.BusinessLogic.csproj (in 239 ms).
[...]

PR #440, opened 2022-05-24
Removed directory separator character and use Path.DirectorySeparatorChar instead

GuimoBear:
Trying to run Mapster.Tool in the docker build process, I came across an error, I believe the error is due to the character that separates the directory hierarchy, the error can be seen in issue #368


andrerav:

cyril265:
I get the same errors as in #368. It's 8.4.0-pre03, right? Not sure if I didn't update it properly, is it fixed for everyone else?

Melethainiel:
> I get the same errors as in #368. It's 8.4.0-pre03, right? Not sure if I didn't update it properly, is it fixed for everyone else?

I still have the errors too

PR #447, opened 2022-06-17
Added ability to skip generating of existing files

psynomorph:
Added the ability to skip recreating and overwriting already existing files by passing a special argument to the tool.


psynomorph:
I'm using Mappster only for generating static mappers and extension methods. Sometimes I need to edit generated mappers to add more complex mapping logic in concrete mapper. Unfortunately Mappster tool rewrites all generated files on each run. In order to generate new mappers, I have to copy the old ones to another location so they don't get overwritten. 
I thought it would be nice to have ability to generate mappers and extensions only for new models and interfaces.

andrerav:
Thank you very much @psynomorph! I have updated the Wiki here to reflect this new functionality: https://github.com/MapsterMapper/Mapster/wiki/Mapster.Tool

PR #449, opened 2022-06-24
Fixes #385 Private Sets

### LINKED ISSUE #385: Private Sets
Hi.

I don't know if it's a feature or a bug, but I have an issue that I don't understand why it's happening. 

I've an Entity with public getters, private setters and a constructor with parameters:

        public class Cargo : Entidade 
        {
         //[...] removed for simplification

        public Cargo(int id, string nome, int? cargoChefeId = null) : base(id)
        {
            _subordinados = new List<Cargo>();
            ValidarNome(nome);
            ValidarCargoChefe(cargoChefeId);
            Nome = nome.Trim().ToUpper();
            CargoChefeId = cargoChefeId;
        }

        public int? CargoChefeId { get; private set; }

        public string Nome { get; private set; }

        public virtual Cargo CargoChefe { get; private set; }

        //[...] removed for simplification
        }

And my View Model:

    public class CargoModelo
    {
        public int Id { get; set; }

        [Display(Name = "Cargo Chefe")]
        public int? CargoChefeId { get; set; }

        [Required]
        [MinLength(5)]
        [MaxLength(100)]
        public string Nome { get; set; }

        public CargoModelo CargoChefe { get; set; }
    }

I wrote this configuration:

         TypeAdapterConfig<CargoModelo, Cargo>.NewConfig().MapToConstructor(true);

It works, but, I don't know why, after mapping to constructor, It uses entity private setters, losing the validation and data manipulation done in the ctor.

How can I change this behavior?

JMPSequeira:
Ensures that if a source property was used in constructor that it will only be reapplied with an explicit resolver.

### DIFF
--- a/src/Mapster.Tests/WhenMappingToConstructorAndPrivateSetters.cs
+++ b/src/Mapster.Tests/WhenMappingToConstructorAndPrivateSetters.cs
@@ -1,0 +1,71 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+
+namespace Mapster.Tests
+{
+    [TestClass]
+    public class WhenMappingToConstructorAndPrivateSetters
+    {
+        [TestInitialize]
+        public void Setup()
+        {
+            TypeAdapterConfig.GlobalSettings.Clear();
+        }
+
+        [TestMethod]
+        public void MapToConstructor_Auto()
+        {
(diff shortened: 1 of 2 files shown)


andrerav:
Solid work, thank you @JMPSequeira!

andrerav:

PR #454, opened 2022-07-18
Refactor IMapper.From to return interface

### LINKED ISSUE #376: Suggestion: IMapper.From should return IAdapterBuilder<TSource> not TypeAdapterBuilder<TSource>
I have currently the following code

private readonly IMapper mapper; // Set by dependency injection

I can easily set up a mockup for IMapper and unit test the following code:

mapper.Map<TDestination>(source);

but how can I test

mapper.From(source).AddParameters("param", 42).AdaptToType<TDestination>();

In my opinion the interface should not return a concrete type, but also an abstraction.

  public interface IMapper
  {
    TypeAdapterConfig Config
    {
      get;
    }

    IAdapterBuilder<TSource> From<TSource>(TSource source); // Not TypeAdapterBuilder<TSource>

    TDestination Map<TDestination>(object source);

    TDestination Map<TSource, TDestination>(TSource source);

    TDestination Map<TSource, TDestination>(TSource source, TDestination destination);

    object Map(object source, Type sourceType, Type destinationType);

    object Map(object source, object destination, Type sourceType, Type destinationType);
  }

gritcsenko:
Refactored `IMapper.From<TSource>()` to return new interface `ITypeAdapterBuilder<TSource>` instead of concrete class.

FYI: Mapster.EF6 & Mapster.EFCore uses outdated package version and causes compilation errors in Mapster.EFCore.Tests and Sample.AspNetCore

Closes #376


andrerav:
@gritcsenko The master branch should be in better shape now with EF Core 6.0. Could you please sync your branch and see if the build errors have gone away? And if so, run the tests as well? Thanks! :)

andrerav:
@gritcsenko I synced the fork and fixed a few build errors. Everything now builds and all tests are passing. Can you take a look at my changes and verify that they are good?

gritcsenko:
> @gritcsenko I synced the fork and fixed a few build errors. Everything now builds and all tests are passing. Can you take a look at my changes and verify that they are good?
@andrerav Thanks for the update. Everything looks good. 
My only concern is that public API was changed

PR #472, opened 2022-08-28
The culture problem has been fixed

mehmetemineker:
During the mapping phase, I encountered a problem caused by the local culture. The `ToLower` and `ToUpper` methods needed to be changed to `ToLowerInvariant` and `ToUpperInvariant`. I fixed this and the unit tests also ran without issue.


andrerav:
@mehmetemineker Thank you very much! :)

PR #397, opened 2021-12-28
Make the project compile and pass tests under net6.0

Philippe-Laval:
Hi all,

I managed to have the Maspter project compile under net6.0 and pass all the unit tests.
I had to add some #if NET6_0_OR_GREATER in the source code.

I created also a project "TemplateTest" to show how to generate code (the T4 templates in the project "Benchmark" are not working, so I had to comment some unit tests for now and take the code of T4 template to the unit tests).

I guess, we could

PR #473, opened 2022-09-05
Update #397 to fix conflicts

andrerav:
This PR is based on the fork at https://github.com/Philippe-Laval/Mapster but updated with the latest changes from master to resolve the current conflicts. Since @Philippe-Laval did not respond on #397 I had to do it in this rather cumbersome way, but it gets the job done.


andrerav:
Note to self: Still need to consider continue using the ExpressionDebugger/ExpressionTranslator nuget packages -- in this PR their respective codebases are added directly as referenced projects. But let's just get this merged first and deal with that later.

Issue #475, opened 2022-09-09
.NET 7 - preview support


danfma:
Hello,

Would be possible to release a version that is compatible with the new .NET 7? I know it's on preview yet but the generator tool fails to load the `System.Runtime` and other assemblies.

Or, maybe, there is a workaround for that... :P

andrerav:
@danfma I think this should be a fairly simple change, and should fit well for the future 7.5 release which will come after the upcoming 7.4 release. Pull requests are extra welcome on this one, but make sure to fork the development branch.

timbze:
I'm not sure about this project, but I had the issue with another library that uses code generation. It was a Roslyn issue, I fixed it by adding this to csproj: `<LangVersion>10.0</LangVersion>`

danfma:
Hello! Ok, I will give it a look later today! 👍🏼

andrerav:
Prerelease packages are now available on NuGet targeting both net7.0 and net6.0. Thank you!

PR #489, opened 2022-10-20
Upgraded to .Net 7

SefaOray:
This PR upgrades all project within the solution to .Net 7. This is created to address the issue: 

https://github.com/MapsterMapper/Mapster/issues/475

### DIFF
--- a/src/Benchmark/Benchmark.csproj
+++ b/src/Benchmark/Benchmark.csproj
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
       <OutputType>Exe</OutputType>
-      <TargetFramework>net6.0</TargetFramework>
+      <TargetFramework>net7.0</TargetFramework>
       <CopyLocalLockFileAssemblies Condition=" '$(Configuration)'=='Debug' ">true</CopyLocalLockFileAssemblies>
       <DefaultItemExcludes>**/*.g.cs</DefaultItemExcludes>
     </PropertyGroup>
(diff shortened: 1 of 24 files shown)


oscarjaergren:
Would love to get a new release with this

cedwards-telis:
This PR still has old references that cause issues with ef 7.0.  Obviously you can use ef core 6.0 with .net 7.0 but you might want to run ef 7.0.

https://github.com/SefaOray/Mapster/blob/development/src/Mapster.EFCore/packages.config

Line 3 should be

  <package id="EntityFramework" version="7.0.0" />

And 

https://github.com/SefaOray/Mapster/blob/development/src/Mapster.EFCore/Mapster.EFCore.csproj

Line 15 should be

    <PackageReference Include="Microsoft.EntityFrameworkCore" Version="7.0.0" />

If you don't do this and run with ef 7.0 you'll get exceptions like

System.MissingMethodException: 'Method not found: 'System.Type Microsoft.EntityFrameworkCore.Metadata.ITypeBase.get_ClrType()'.' due to conflicting versions.

I suppose that there is a way to handle complex mixed version packages but I don't know it....

HenricFranssonXM:
Do you have a release plan for official .Net 7 support?

cedwards-telis:
Actually I think the issue I am having with

System.MissingMethodException: 'Method not found: 'System.Type Microsoft.EntityFrameworkCore.Metadata.ITypeBase.get_ClrType()'.' due to conflicting versions.

may be because the verison of the ef core plugin on Nuget was not complied on the right version of ef core - https://github.com/dotnet/efcore/issues/26701#issuecomment-1054881574

andrerav:
Fixed by #496. Thank you @SefaOray!

Issue #337, opened 2021-05-14
Include #nullable enable to generated files


scottcollins:
I have some models being generated in a project that uses nullable reference types

The types generate fine, but I am getting CS8669 warnings because the generated file does not have #nullable enable included

Ex: `public string? City { get; set; }` gets generated but is marked as a warning

Adding this directive to the generated file would eliminate these warnings

SML-DmitrySokolov:
Attempt to avoid this warning by changing the type from "nullable"(string?) to  "non nullable"(string) also  has not effect. Looks like Mapster does not see the difference between string? and string. 

About changing a type:
https://github.com/MapsterMapper/Mapster/wiki/Fluent-API-Code-generation#change-a-property-name-type

andrerav:
@scottcollins Is it possible for you to solve this by adding `<Nullable>enable</Nullable>` in the csproj that contains the generated files?

cyril265:
The warning says `PersonDto.g.cs(12, 22): [CS8669] The annotation for nullable reference types should only be used in code within a '#nullable' annotations context. Auto-generated code requires an explicit '#nullable' directive in source.`

Apparently it still needs `#nullable enable` in the file even with `<Nullable>enable</Nullable>` in the project.

Elgirhath:
Having the same problem. Any update on this? It's a blocker for us

Elgirhath:
As noted in Microsoft's documentation [here in an `Important` section](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references#nullable-contexts) - generated files need to include explicit `#nullable enable` line.

inmhoel:
I have experienced the same. Some more info can be found here: https://stackoverflow.com/a/59460631

As @cyril265 says, `<Nullable>enable</Nullable>` in csproj is ignored - Roslyn requires an explicit `nullable` directive in generated code files (*.**g**.cs).

kaddyuk:
+1 Getting this warning. And with `<WarningAsErrors>nullable</WarningAsErrors>`, the tool breaks the build.

oscarjaergren:
I would like for this issue to get some attention too, as it's a pain to have to manually add it back to each generated DTO as soon as I rebuild.

cyril265:
@andrerav I'm just going to mention you here, not sure if you are subscribed to this issue.

PR #530, opened 2023-01-24
Add options to Mapster.Tool to generate files with #nullable directives

kescherCode:
Fixes https://github.com/MapsterMapper/Mapster/issues/337 for Mapster.Tool

This allows to add "--nullableDirective" to the options already set within a given csproj file to add an `#nullable enable`, in order to suppress CS8669 warnings that are raised specifically due to generated source files having to add this directive even if a project has `Nullable` enabled.


andrerav:
Hi @kescherCode, thank you for contributing this, this has been requested a few times now so I think you will make a lot of people happy :) Would you consider adding a short form option as well? The obvious choice (-n) is already taken, but perhaps -N will work?

andrerav:
Awesome, thanks!

PR #533, opened 2023-02-02
Add IMapFrom interface for automatic mapping configuration 

stormaref:
Creating IMapFrom interface
now destination type can inherit from an interface named IMapFrom and they can configure mapping in the same class like this:
```csharp
public class InheritedDestinationModel : IMapFrom<SourceModel>
{
    public string Type { get; set; }
    public int Value { get; set; }

    {
        config.NewConfig<SourceModel, InheritedDestinationModel>()
            .Map(dest => dest.Value, _ => DesireValues.Number);
    }
}
```

and even if you don't implement the method it will automatically create a new config for desire types
ex:
```csharp
public class DestinationModel : IMapFrom<SourceModel>
{
    public string Type { get; set; }
}
```
I've added a new extension method to TypeAdapterConfig named ScanInheritedTypes like this:
```csharp
TypeAdapterConfig.GlobalSettings.ScanInheritedTypes(Assembly.GetExecutingAssembly());
```

I've created tests for all situations and tests are passed
as soon as this pull request merged I will update wiki page

### DIFF
--- a/src/Mapster.Tests/WhenUsingIMapFrom.cs
+++ b/src/Mapster.Tests/WhenUsingIMapFrom.cs
@@ -1,0 +1,69 @@
+using System.Reflection;
+using Mapster.Utils;
+using MapsterMapper;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+
+namespace Mapster.Tests;
+
+[TestClass]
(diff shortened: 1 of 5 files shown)


andrerav:
Thanks @stormaref!

Issue #513, opened 2022-12-28
Mapster Tool Mapper crashes when using MediatR (and possibly other assemblies)


Christoph-Wolf-Coduct:
PS C:\Users\Chris\source\repos\JobScrape.Backend> dotnet mapster mapper -a JobScrape.Backend\obj\Debug\net6.0\JobScrape.Backend.dll

Produces this Exception: 

Unhandled exception. System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types.
Could not load file or assembly 'MediatR.Contracts, Version=1.0.1.0, Culture=neutral, PublicKeyToken=bb9a41a5e8aaa7e2'. Invalid pointer (0x80004003 (E_POINTER))
Could not load file or assembly 'MediatR, Version=11.0.0.0, Culture=neutral, PublicKeyToken=bb9a41a5e8aaa7e2'. Invalid pointer (0x80004003 (E_POINTER))
   at System.Reflection.RuntimeModule.GetTypes(RuntimeModule module)
   at System.Reflection.Assembly.GetTypes()
   at Mapster.TypeAdapterConfig.<>c.<Scan>b__87_0(Assembly assembly) in C:\Projects\Mapster\src\Mapster\TypeAdapterConfig.cs:line 639
   at System.Linq.Enumerable.SelectArrayIterator`2.MoveNext()
   at System.Linq.Enumerable.SelectManySingleSelectorIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Mapster.TypeAdapterConfig.Scan(Assembly[] assemblies) in C:\Projects\Mapster\src\Mapster\TypeAdapterConfig.cs:line 639
   at Mapster.Tool.Program.GenerateMappers(MapperOptions opt) in C:\Projects\Mapster\src\Mapster.Tool\Program.cs:line 65
   at CommandLine.ParserResultExtensions.WithParsed[T](ParserResult`1 result, Action`1 action)
   at Mapster.Tool.Program.Main(String[] args) in C:\Projects\Mapster\src\Mapster.Tool\Program.cs:line 17
System.IO.FileLoadException: Could not load file or assembly 'MediatR.Contracts, Version=1.0.1.0, Culture=neutral, PublicKeyToken=bb9a41a5e8aaa7e2'. Invalid pointer (0x80004003 (E_POINTER))
[...]

Issue #467, opened 2022-08-10
Mapster.Tool crashes when using WPF


Irme:
I am trying to generate a mapper using Mapster.Tool but it crashes as soon as WPF is added as a dependency.

Creating a simple console app for .NET6.0-windows works fine, I can generate DTOs and Mappers as I need. Whenever I use essentially the same code, but add it to a WPF project the mapster build targets cause an error, and no code is generated.

I've uploade a [MWE to github](https://github.com/Irme/MappingTesting), and the error is pasted below

```
Build started...
1>------ Build started: Project: MappingTestWPFApp, Configuration: Debug Any CPU ------
1>Skipping analyzers to speed up the build. You can execute 'Build' or 'Rebuild' command to run analyzers.
1>Skipping analyzers to speed up the build. You can execute 'Build' or 'Rebuild' command to run analyzers.
1>MappingTestWPFApp -> C:\Users\irme\source\repos\MappingTesting\MappingTestWPFApp\bin\Debug\net6.0-windows\MappingTestWPFApp.dll
1>Tool 'mapster.tool' (version '8.3.0') was restored. Available commands: dotnet-mapster
1>
1>Restore was successful.
1>Cannot find library: PresentationFramework
1>Cannot find library: PresentationFramework
1>Unhandled exception. System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types.
1>Could not load file or assembly 'PresentationFramework, Version=6.0.2.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
1>Could not load file or assembly 'PresentationFramework, Version=6.0.2.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.
1>   at System.Reflection.RuntimeModule.GetTypes(RuntimeModule module)
1>   at System.Reflection.Assembly.GetTypes()
1>   at Mapster.Tool.Extensions.Scan(CodeGenerationConfig config, Assembly assembly) in C:\Projects\Mapster\src\Mapster.Tool\Extensions.cs:line 177
1>   at Mapster.Tool.Program.GenerateModels(ModelOptions opt) in C:\Projects\Mapster\src\Mapster.Tool\Program.cs:line 146
1>   at CommandLine.ParserResultExtensions.WithParsed[T](ParserResult`1 result, Action`1 action)
[...]

Issue #455, opened 2022-07-22
Running Mapper Generation in Docker


doddi321:
I'm trying to use the generation tool to generate my mapper interfaces but every time I try to build my docker image I get an error.

System.IO.FileNotFoundException

#15 11.31   Restored /src/Application/Application.csproj (in 9.61 sec).
#15 11.31   Restored /src/WebApplication1/WebApplication1.csproj (in 9.61 sec).
#15 13.12   Application -> /app/build/Application.dll
#15 15.27   Tool 'mapster.tool' (version '8.3.0') was restored. Available commands: dotnet-mapster
#15 15.27
#15 15.27   Restore was successful.
#15 15.70   Cannot find assembly path: FluentValidation (type=package, version=11.1.0)
#15 15.71   Unhandled exception. System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types.
#15 15.71   Could not load file or assembly 'FluentValidation, Version=11.0.0.0, Culture=neutral, PublicKeyToken=7de548da2fbae0f0'. The system cannot find the file specified.
#15 15.71
#15 15.71      at System.Reflection.RuntimeModule.GetTypes(RuntimeModule module)
#15 15.71      at System.Reflection.Assembly.GetTypes()
#15 15.71      at Mapster.Tool.Extensions.Scan(CodeGenerationConfig config, Assembly assembly) in C:\Projects\Mapster\src\Mapster.Tool\Extensions.cs:line 177
#15 15.71      at Mapster.Tool.Program.GenerateModels(ModelOptions opt) in C:\Projects\Mapster\src\Mapster.Tool\Program.cs:line 146
#15 15.71      at CommandLine.ParserResultExtensions.WithParsed[T](ParserResult`1 result, Action`1 action)
#15 15.71      at Mapster.Tool.Program.Main(String[] args) in C:\Projects\Mapster\src\Mapster.Tool\Program.cs:line 17
#15 15.71   System.IO.FileNotFoundException: Could not load file or assembly 'FluentValidation, Version=11.0.0.0, Culture=neutral, PublicKeyToken=7de548da2fbae0f0'. The system cannot find the file specified.
#15 15.71
#15 15.71   File name: 'FluentValidation, Version=11.0.0.0, Culture=neutral, PublicKeyToken=7de548da2fbae0f0'
[...]

PR #544, opened 2023-02-14
Mapster.Tool catch ReflectionTypeLoadException

EniacMlezi:
Was running into Mapster.Tool reflection issues. Docker debug builds (Fastmode enabled, windows build) were succeeding, but release builds (Fastmode disabled, linux build) were failing. Don't know exactly why ReflectionTypeLoadException was occuring when loading certain types from an Assembly, but mr Jon Skeet gave a beautiful (/s) piece of code to filter these exceptions out and continue with all types that loaded successfully: https://stackoverflow.com/a/7889272. 

I'm afraid this also might filter out IRegister Types, but I'm not too familiar with the way `Assembly.Load()` and `.GetTypes()` work.

Possibly fixes:
#368
#467
#455
#513


src/Mapster.Core/Utils/Extensions.cs
```diff
@@ -1,9 +1,12 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 
 namespace Mapster.Utils
 {
-    static class Extensions
+    public static class Extensions
```
EniacMlezi, line 9: Made this public to use from Mapster and Mapster.Tool. Maybe not ideal if it gets published?

andrerav: Looks fine to me to be honest, nothing problematic being exposed in that class.

andrerav:
Thank you!

Issue #536, opened 2023-02-07
Mapping with existing object and init properties are failed


YaroslavMudryk:
I have object User and UserDto:

```C#
public class User
{
   public int Id { get; init; }
   public string Name { get; init; }
}

public class UserDto
{
   public int Id { get; init; }
   public string Name { get; init; }
}
```
Also i have config:
```C#
public class UserMappingRegister : IRegister
{
    public void Register(TypeAdapterConfig config)
    {
        config.NewConfig<User, UserDto>()
            .MapToConstructor(true)
            .ConstructUsing(s => new UserDto());
    }
}

```
And interface for mappers:
```C#
[Mapper]
public interface IUserMapper
{
    Expression<Func<User, UserDto>> UserProjection { get; }
    UserDto MapTo(User user);
    UserDto MapTo(User user, UserDto userDto);
}
```
As result I get the generated file:
```C#
public partial class UserMapper: IUserMapper
{
    public Expression<Func<User, UserDto>> UserProjection => p1 => new UserDto()
    {
        Id = p1.Id,
        Name = p1.Name
    }
    public UserDto MapTo(User p2)
    {
        return p2 == null ? null : new UserDto()
        {
            Id = p2.Id,
            Name = p2.Name
        };
    }
    public UserDto MapTo(User p3, UserDto p4)
    {
        if (p3 == null)
        {
            return null;
        }
        UserDto result = p4 ?? new UserDto();
            
        result.Id = p3.Id; //there is an error due to property marked as init
        result.Name = p3.Name; //there is an error due to property marked as init
        return result;
    }
}
```

andrerav:
This scenario is not supported by Mapster.Tool at the moment. PR's are welcome.

YaroslavMudryk:
Ok, thanks a lot!

stormaref:
I'm working on it

PR #545, opened 2023-02-16
Fix Issues #536 -> Set properties with reflection if destination has init only properties.

stormaref:
I've fixed [#536](https://github.com/MapsterMapper/Mapster/issues/536) and wrote a test that regenerates the exact issue
The test has been passed
I add a method to "ClassAdapter" in [this commit](https://github.com/MapsterMapper/Mapster/commit/f3df57c56841cf6230727f5e2ddf5d73e0c33fa8) to generate reflection code for setting init only properties

The result is like this:
```csharp
public _UserDto MapTo(_User p3, _UserDto p4)
        {
            if (p3 == null)
            {
                return null;
            }
            _UserDto result = p4 ?? new _UserDto();
            
            typeof(_UserDto).GetProperty("Id").SetValue(result, (object)p3.Id);
            typeof(_UserDto).GetProperty("Name").SetValue(result, p3.Name);
            return result;
            
        }
```
* I did not update the package version in csproj

### DIFF
--- a/src/Mapster.Tool.Tests/.config/dotnet-tools.json
+++ b/src/Mapster.Tool.Tests/.config/dotnet-tools.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "mapster.tool": {
+      "version": "8.3.0",
+      "commands": [
+        "dotnet-mapster"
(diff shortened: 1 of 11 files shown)


andrerav:
Thank you!

PR #579, opened 2023-04-15
Remove mentions to old unsupported frameworks from projects

satano:
- Cleanup of conditions in project files to other (old) frameworks, which are not used anymore. All projects in solution targets just .NET 6.
- I also changed element `<TargetFrameworks>` to `<TargetFramework>` everywhere.

### DIFF
--- a/src/ExpressionDebugger/ExpressionDebugger.csproj
+++ b/src/ExpressionDebugger/ExpressionDebugger.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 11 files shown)


satano:
I made this PR because there really is no need for these conditions when those frameworks are not supported. But I am curious, Mapster is dropping support for (old) .NET Framework? I do not mean .NET Core 3.x (or less), but full .NET Framework 4.x.

andrerav:
Thanks! Yes, it's not sustainable to maintain support for .NET Framework unfortunately (this has to do with how the code generation works). Users that require this library on that platform can use the older versions of Mapster.

PR #583, opened 2023-04-22
🐛 fix(TypeAdapter.cs): Make source parameter nullable

msadeqsirjani:
# Changes
* The source parameter in the Adapt method is now nullable, which allows for null values to be passed in without throwing a null reference exception. This improves the robustness of the code and makes it more defensive against unexpected null values.

@eswann @chaowlert @chaowlert @satano


andrerav:
Thanks!

PR #586, opened 2023-05-05
Generating valid, meaningful method name from destination type

### LINKED ISSUE #580: Mapster.Tool generates incorrect mapping extension method names when a destination type is nullable or generic
Mapster.Tool generates incorrect mapping extension method names when a destination type is nullable or generic

**Example:**
Declaration:
```c#
    public class GenericTestSrc
    {
        public int Val { get; set; }
    }
    public class GenericTestDst<T>
    {
        public T Val { get; set; }
    }

    public record MyDateTime(DateTime? DateTime);
```

Mapper config:
```c#

            config.ForType<MyDateTime, DateTime?>()
                .MapWith(s => s.DateTime.Value )
                .GenerateMapper(MapType.Map);

            config.ForType<GenericTestSrc, GenericTestDst<int>>()
                .GenerateMapper(MapType.Map);
```

What gets generated:
```c#
        public static System.DateTime? AdaptToDateTime?(this MapLib.MapTest.MyDateTime s)
        {
            return (System.DateTime?)s.DateTime.Value;
        }

        public static MapLib.MapTest.GenericTestDst<int> AdaptToGenericTestDst<int>(this MapLib.MapTest.GenericTestSrc p1)
        {
            return p1 == null ? null : new MapLib.MapTest.GenericTestDst<int>() {Val = p1.Val};
        }
```

Attached sample project:
[MapsterBug.zip](https://github.com/MapsterMapper/Mapster/files/11285014/MapsterBug.zip)

rafalka:
This PR fixes #580

Examples:
```
typeof(int) => Int
typeof(int?) => Int
typeof(int?[]) => IntArray
typeof(List<string[]>) => StringArrayList
```

### DIFF
--- a/src/Mapster.Tool/Program.cs
+++ b/src/Mapster.Tool/Program.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Text;
 using CommandLine;
 using ExpressionDebugger;
 using Mapster.Models;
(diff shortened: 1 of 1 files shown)


andrerav:
Thanks!

Issue #370, opened 2021-09-03
Properties that contain capitalized abbreviations cannot also be mapped in the fluent API when mapping to a record


CerebralUnit:
There are several properties in the Adverse record on the domain model that have abbreviations and are therefore capitalized (e.g. `HDMAReasons.`) Regardless of the naming strategy Exact, IgnoreCase, Flexible etc.. Mapster will fail execute a custom mapping unless a Pascal case version of the property is passed in for the destination object. Using x => x.HDMAReasons causes the mapper to throw a configuration error saying that there is neither a mapping nor an Ignore for `HdmaReasons.` Neither object being mapped has that casing. Example of the code that finally worked below. 

config.MapDependentTo<Adverse, Domain.Adverse>()
                  .MapToConstructor(true)
                  .NameMatchingStrategy(NameMatchingStrategy.IgnoreCase)
                  .Map("HmdaReasons", src => src.HMDAReasons.Select(x => x.EnumValue))

andrerav:
Might be related to #388.

JMPSequeira:
@CerebralUnit could you provide the models used for the issue? And please clarify if the specific problem only occurs when mapping to a constructor.

ventii:
Had the exact same issue - fields with consecutive capital letters in a C# record constructor are not mapped even when the source record/class has the same field with the same name.

non working example:
```
public record MySourceRecord(DateTime ValidFromUTC);

public record MyDestinationRecord(DateTime ValidFromUTC);

// having the following config
config.NewConfig<MySourceRecord, MyDestinationRecord>();
```

got it to work by doing the following changes:

```
// change property name of destination record to PascalCase
public record MyDestinationRecord(DateTime ValidFromUtc);

// update config to specify mapping

config.NewConfig<MySourceRecord, MyDestinationRecord>()
                .Map(dest => dest.ValidFromUtc, src => src.ValidFromUTC);
```
I was going crazy as I initially thought it was some bug with DateTime. 

Is this a known issue please?

andrerav:
@ventii Yes this is a known issue, the name matching strategy does not seem to be working as intended.

andrerav:
See also #388.

leseneda:
I solved my issue using the following code line;

globalSettings.Default.NameMatchingStrategy(NameMatchingStrategy.Flexible);

Mapster 7.4.0

Hope this helps.

Berkay-Sari:
[...]

PR #590, opened 2023-05-21
Fix MapsterMapper#370 where capitalized abbreviations are not mapped …

joshslark:
…properly between records

This PR fixes an issue where when you have a destination member that isn't in pascal case for a record, then it will ignore the mapping. (#370) 

In `ValueAccessingStrategy.cs`, the custom resolver destination member is compared for equality against the ParameterModel destination member. However, in ParameterModel, it modifies the destination member to force it to be Pascal Case. I couldn't figure out exactly why we do this, but it breaks a number of tests if you change it. 

The only solution I could find that avoids breaking any tests is to compare the two destination members ignoring case. Even if this isn't the correct solution, hopefully this will help someone more familiar create a better solution.

### DIFF
--- a/src/Mapster.Tests/WhenMappingRecordTypes.cs
+++ b/src/Mapster.Tests/WhenMappingRecordTypes.cs
@@ -37,6 +37,19 @@ namespace Mapster.Tests
             dest.Age.ShouldBe(10);
         }
 
+        [TestMethod]
+        public void Map_RecordType_CapitalizationChanged()
+        {
+            TypeAdapterConfig<RecordType, RecordTypeDto>.NewConfig()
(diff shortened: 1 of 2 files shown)


RobKoda:
Hello!

Any idea on when this will be published? Even as a prerelease nuget? 
Latest prerelease (pre06) seems to be 7 months old, while this fix was merged 4 months ago.

andrerav:
Hi @RobKoda, I released 7.4.0 yesterday. Apologies for the delay, it's been a busy summer :)

elmondir:
this still happen with property Name FrName

mehmetozkaya:
Error still continue and not mapped _ properties i.e. Address_FirstName in 7.4.0 version.

PR #606, opened 2023-07-05
Added types used in rules to types hashset

### LINKED ISSUE #605: Mapster.Tool: Mapper code is not generated when source type is a generic
Mapster.Tool does not generate mapper class when source type is a generic
### Types definition:
```C#
public enum TypeEnum { A, B, C }

public class Src<T> where T:Enum
{
    public int IntVal { get; set; }
    public T Type { get; set; }
}

public class Dst<T> where T:Enum
{
    public int IntVal { get; set; }
    public T Type { get; set; }
}
```

### Mapper configuration:
```C#
        {
            config.NewConfig<Src<TypeEnum>, Dst<TypeEnum>>()
                .GenerateMapper(MapType.Map);
        }
```

### Sample project:
[MapsterBug-CodeNotGeneratedForGeneric.zip](https://github.com/MapsterMapper/Mapster/files/11955281/MapsterBug-CodeNotGeneratedForGeneric.zip)

rafalka:
This PR fixes #605 

Assembly defines open generic only, so we had to add specialised types used in mappings
 
Removed removing "Nullable" from type name in order to prevent method/class name collisions

### DIFF
--- a/src/Mapster.Tool/Program.cs
+++ b/src/Mapster.Tool/Program.cs
@@ -80,7 +80,7 @@ namespace Mapster.Tool
                 {
                     Implements = new[] {type},
                     Namespace = CreateNamespace(opt.Namespace, segments, type.Namespace),
-                    TypeName = attr.Name ?? GetImplName(type.Name),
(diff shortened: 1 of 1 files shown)


andrerav:
Thank you!

PR #629, opened 2023-09-18
Not using shallow copy if mapping rule exists

### LINKED ISSUE #628: `MapWith()` and `MapToTargetWith()` are ignored when `ShallowCopyForSameType` is set
When `ShallowCopyForSameType` then custom conversion defined by `MapWith()` or `MapToTargetWith()` is ignored

## Sample code (based on existing UT with added `ShallowCopyForSameType` config)
```C#
      [TestMethod]
      public void MapToTargetWith_Should_Work_With_Adapt_To_Target()
      {
          var a = new Foo { A = 1 };
          var b = new Bar { A = 2 };

          //This will not work as expected => b.A will be 1, ignoring the mapping defined
          config.Default.ShallowCopyForSameType(true);

          config.NewConfig<double, double>().MapToTargetWith((x, y) => 5);
          a.Adapt(b, config);
          b.A.ShouldBe(5);
      }

      internal class Foo
      {
          public double A { get; set; }
      }

      internal class Bar
      {
          public double A { get; set; }
      }
```

rafalka:
This PR fixes #628 

Added checking if a mapping conversion is defined for desired type

### DIFF
--- a/src/Mapster.Tests/WhenSettingMapToTargetWith.cs
+++ b/src/Mapster.Tests/WhenSettingMapToTargetWith.cs
@@ -17,11 +17,38 @@ namespace Mapster.Tests
             var a = new Foo { A = 1 };
             var b = new Bar { A = 2 };
 
-            //This will not work as expected => b.A will be 1, ignoring the mapping defined
             var config = new TypeAdapterConfig();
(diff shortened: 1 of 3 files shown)


andrerav:
Thank you!

Issue #640, opened 2023-10-08
Mapping nullable Enum to custom object


WahidBitar:
for now I can have something like the following to register mapping for all enums in the system
```
config.NewConfig<Enum, KeyValueData>()
            .MapWith(s => new KeyValueData(s, Enums.ResourceManager));
```

However, it's not possible to have the same cofig for a nullable Enum. 

The following config is not working:

```
config.NewConfig<Enum?, KeyValueData?>()
            .MapWith(s => s == null ? null : new KeyValueData(s.ToString(), Enums.ResourceManager));
```

I'll get the following Exception

Error while compiling
source=Tests.FirstClass
destination=Tests.SecondClass
type=Map
-> InnerException:
Error while compiling
source=System.Nullable`1[Model.IdentityType]
destination=Model.KeyValueData
type=Map
--> InnerException:
No default constructor for type 'KeyValueData', please use 'ConstructUsing' or 'MapWith'

WahidBitar:
I've figured a way out but still think it might be a better if we've somthing more helpful in the config. Especially from primitive types to classes. Anyways, my solution was simply like the following:

```
config.NewConfig(typeof(Nullable<>), typeof(KeyValueData))
            .IgnoreNullValues(true)
            .Map("key", "Value")
            .MapToConstructor(true);
```

DocSvartz:
Hello @WahidBitar , if your goal is to skipped mapping when the Source is null, maybe you help [.IgnorIf()] or (https://github.com/MapsterMapper/Mapster/wiki/Ignoring-members#ignore-conditionally)

DocSvartz:
Hello @WahidBitar, I ran into this problem while preparing a fix for the problem record detection. 

I think the problem is as follows: Nullable<T> in Mapster terminology should adapt as Record - forced creation via constructor. But now it happens as a Class where without special directives, there is only creation through a constructor without parameters.

WahidBitar:
Thanks @DocSvartz , My main goal is not to skip the null values. My goal is to map from a nullable enum to my custom class `KeyValueData` in a general way.
I'll put here the entire aspects and my work around so you may take benifit of it.

Note: you may found something weard in my `KeyValueData` class because I have a general mapper from Enum to string because I put translation of the enums in the Resources files, so If you don't have this case your config should be simpler. Anyway here is my classes and config:

Mapping config:
```
public class GeneralMapper : IRegister
{
    public void Register(TypeAdapterConfig config)
    {
[...]

PR #643, opened 2023-10-15
Fix issue #640 - Nullable enum to nullable destination (class or param)

DocSvartz:
Fix from Isuue #640

It Will Work When Enum Cast to object.  I'm not sure this is the right way.


andrerav:
Closed and reopened to re-run the workflow.

andrerav:
Thank you @DocSvartz!

PR #645, opened 2023-10-18
Fix issue #524 - Add support for working with types packed into an object to the standard adapter

### LINKED ISSUE #524: Invalid mapping to exists class if source is object
```csharp
public class Source
{
    public int X1 { get; set; }
}
public class Dest
{
    public int X1 { get; set; }
}

var source = new Source { X1 = 123 };
var dest = new Dest { X1 = 321 };
var dest1 = source.Adapt(dest);
Console.WriteLine($"{dest.X1} {dest1.X1} {dest.GetHashCode()} {dest1.GetHashCode()}");
```
result: `123 123 19575591 19575591`, dest and dest1 are the same object

```csharp
void somemap(object source)
{
    var dest = new Dest { X1 = 321 };
    var dest1 = source.Adapt(dest);

    Console.WriteLine($"{dest.X1} {dest1.X1} {dest.GetHashCode()} {dest1.GetHashCode()}");
}

var source = new Source { X1 = 123 };
somemap(source);
```
result: `321 123 32001227 19575591`
dest not changed, dest1 is a new object.

Any ideas? Why if source is object, Adapt create a new instance of dest class?

DocSvartz:
Fix issue #524

Before:

 if Type was packaged into an object:

`object _source = new TSource()
`

Instead of updating with data from TSource, it was converted to the TDestination type

`_source.Adapt(_destination) == _source.Adapt<TDistination>`

### DIFF
--- a/src/Mapster.Tests/WhenMappingObjectRegression.cs
+++ b/src/Mapster.Tests/WhenMappingObjectRegression.cs
@@ -1,0 +1,156 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+using System;
+
+namespace Mapster.Tests
+{
+    [TestClass]
+    public class WhenMappingObjectRegression
+    {
+        /// <summary>
+        /// https://github.com/MapsterMapper/Mapster/issues/524 
+        /// </summary>
+        [TestMethod]
+        public void TSourceIsObjectUpdate()
+        {
+            var source = new Source524 { X1 = 123 };
+            var _result = Somemap(source);
+
+            _result.X1.ShouldBe(123);
+        }
+
+        /// <summary>
+        /// https://github.com/MapsterMapper/Mapster/issues/524
+        /// </summary>
+        [TestMethod]
+        public void TSourceIsObjectUpdateUseDynamicCast()
+        {
+            var source = new Source524 { X1 = 123 };
+            var _result = SomemapWithDynamic(source);
+
+            _result.X1.ShouldBe(123);
+        }
+
+        [TestMethod]
+        public void UpdateManyDest()
+        {
+            var source = new Source524 { X1 = 123 };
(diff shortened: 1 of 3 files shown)


DocSvartz:
@andrerav  It still doesn't work as it should. 

Now it working  this  `_source.Adapt<TDestination>().Adapt(TDestination);`

There will be problems with updating when the number of fields in the TSource is less than in TDestination

DocSvartz:
Seems like, It was simply necessary to apply an already developed solution. By adding it to a standard adapter. :)

DocSvartz:
Hello @andrerav  
It looks like the simplest and most correct solution would actually be to use the mechanism from the special overload of the Adapt Method.
1) In general, the Expression for an adapt is created [based on the types of generic arguments.](https://github.com/MapsterMapper/Mapster/blob/04ac871b55828c3909b6cee4764e6fab40db3983/src/Mapster/TypeAdapter.cs#L96)
With this call, generation will always occur for the object type as a TSource or TDestination. Since the generic method was called for them :)
2) The Received conversion Adapt function is [ called with Runtime arguments (variables) ](https://github.com/MapsterMapper/Mapster/blob/04ac871b55828c3909b6cee4764e6fab40db3983/src/Mapster/TypeAdapter.cs#L97) passed for updating.
But since the lambda has already been generated for the Object type, processing for it will occur as an Object.

Therefore, this is not exactly an adapter object problem, as I initially decided.
You just need to initially generate the Adapt function for the type Packed into an object and not the object itself

DocSvartz:
@andrerav  It looks like now the work on the fix is really complete.
I left the endless loop breaker in both places, for greater reliability ))

andrerav:
@DocSvartz Could you please resolve this merge conflict when you have time? Looks like it should be fairly trivial :)

andrerav:
Thank you @DocSvartz!

Issue #537, opened 2023-02-07
Critical regression introduced in Mapster 7.3.0 when mapping to existing objects


andrerav:
There are numerous issues that have been reported since the release of 7.3.0 that seem to have a common theme -- performing a map to an existing object has a tendency to create a new object instance instead of preserving the instance being mapped. So far it seems that the following issues have reported this:

- #427 
- #430
- #498 
- #485
- #482 
- #524 

Other possibly related issues:
- #546

I'm working on investigating which chance in 7.3.0 that might have introduced this issue. I will update this issue when I have new information to share.

satano:
Hello.

In this case, should not be version 7.3.0 marked as deprecated because of critical bugs in [nuget.org](https://www.nuget.org/packages/Mapster/7.3.0)?

doddgu:
hi, is there any progress?

DocSvartz:
Hello, new instanse always created if TDestination does not have a public constructor without parameters. Maybe this has something to do with the problem.

andrerav:
Thanks to @DocSvartz this issue has been partially fixed. See #634 for details.

DocSvartz:
Hello @andrerav , have you had problems with dictionaries ?
The generic KeyValuePair has always been treated as a record.

DocSvartz:
@andrerav  I open PR #646  When  2 Step of Fixing this problemm :)

andrerav:
Hopefully fixed by #649, #646 and #634. Prerelease packages are available on NuGet as of this writing.

Issue #501, opened 2022-11-29
Multiple sources - Mapping required for every member


ChristopheLionet:
Issue https://github.com/MapsterMapper/Mapster/issues/180 shows how to solve the need for multiple sources using tuples, and that it is implemented in Mapster 5.0.

However, using multiple sources as stated in the docs introduces a heavy constraint, which is to map every member manually.
In the following code, dto.CommonValue is null if not explicitly mapped in the config.

```csharp
record PocoA(string SpecificValue);      
record PocoB(int CommonValue);                
record Dto(int CommonValue, int SumValue);  
//...

TypeAdapterConfig.GlobalSettings.NewConfig<(PocoA, PocoB), Dto>()                                                 
    .Map(dto => dto.SumValue, pocoTuple => pocoTuple.Item1.SpecificValue + pocoTuple.Item2.CommonValue);    
// ...

var dto = mapper.Map<(PocoA, PocoB), Dto>(pocoA, pocoB);  // dto.CommonValue is null
```

Is there a solution to this?
It breaks the workflow in some large objects. If a single parameter needs an additional source, every single member needs to be mapped manually.

ocdi:
I too just suffered from this, the doc here doesn't seem to work https://github.com/MapsterMapper/Mapster/wiki/Custom-mapping#multiple-sources

I get an exception when I try to use the Example 2

```
System.ArgumentException: Allow only member access (eg. obj => obj.Child.Name) (Parameter 'lambda')
```

If I manually map each, it works. But that defeats the point of mapping. :-)

andrerav:
Hi @ChristopheLionet, could you post a complete code sample please?

jackfrosty908:
I am also having this issue, though maybe I am not setting up the config correctly? 

Code can be run [here](https://dotnetfiddle.net/MOvzYp)

```
using System;
using Mapster;
					
public class Program
{
	public static void Main()
	{
		
		TypeAdapterConfig<(Source1, Source2), Dest>.NewConfig()
                .Map(dest => dest, src => src.Item1)
                .Map(dest => dest, src => src.Item2);
		
		var src1 = new Source1{MyInt = 1};
		var src2 = new Source2{MyString = "Hello"};
		
		//Doesnt work
		var destTuple = Tuple.Create(src1, src2).Adapt<Dest>();
		
		//Works
		var destSrc1 = src1.Adapt<Dest>();
		var destSrc2 = src2.Adapt<Dest>();
		
[...]

PR #646, opened 2023-10-23
Fix issue #537 - Step 2 - Add support Extended Record Function + Attribute from support used current record definitions

DocSvartz:
Fix Issue:
#501 - _test named OnlyInlineRecordWorked_ (regarding not filling the record with data in the given case)

Add 
1) (InlineExpression is Worked) Field Activation passing Field values not specified in the constructor when creating a new record instance
2) Record Creation on Default Passing Field values not specified in the constructor when creating a new record instance (Activation)
3) A special attribute to support custom classes (structures) that use the [current record definition](https://github.com/MapsterMapper/Mapster/wiki/Data-types#mappable-objects). _description  of spec record at the end of the page_ -
[AdaptWith(AdaptDirectives.DestinationAsRecord)] - **to be confirmed**

Example:
```
 [TestMethod]
        public void MultiCtorAndInlineRecordWorked()
        {
            var _sourcePoco = new MultiCtorAndInlinePoco() { MyInt = 1, MyString = "Hello", MyEmail = "123@gmail.com", InitData="Test"};
            var _sourceMultiCtorAndInline = new MultiCtorAndInlineRecord(2, "Hello World") { InitData = "Worked", MyEmail = "243@gmail.com" };

            var _resultMultiCtorAndInline = _sourcePoco.Adapt<MultiCtorAndInlineRecord>();  // Activation 
            var _updateResult = _sourceMultiCtorAndInline.Adapt(_resultMultiCtorAndInline);  // Update

            _resultMultiCtorAndInline.MyInt.ShouldBe(1);
            _resultMultiCtorAndInline.MyString.ShouldBe("Hello");
            _resultMultiCtorAndInline.MyEmail.ShouldBe("123@gmail.com");
            _resultMultiCtorAndInline.InitData.ShouldBe("Test");
            _updateResult.MyInt.ShouldBe(2);
            _updateResult.MyString.ShouldBe("Hello World");
            _updateResult.MyEmail.ShouldBe("243@gmail.com");
            _updateResult.InitData.ShouldBe("Worked");
        }

    record MultiCtorAndInlineRecord
    {
        public MultiCtorAndInlineRecord(int myInt)
        {
            MyInt = myInt;
        }

        public MultiCtorAndInlineRecord(int myInt, string myString) : this(myInt) // default have been using this constructor
        {
            MyString = myString;
        }


        public int MyInt { get; private set; } // private set only constructor or using init 
        public string MyString { get; private set; } // private set only constructor or using init 
        public string MyEmail { get; set; } 
        public string InitData { get; init; } // init in Constructor or inline activation
    }

```

### DIFF
--- a/src/Mapster.Core/Attributes/AdaptAsAttribute.cs
+++ b/src/Mapster.Core/Attributes/AdaptAsAttribute.cs
@@ -1,0 +1,18 @@
+﻿using System;
+
+namespace Mapster.Attributes
+{
+    [AttributeUsage(AttributeTargets.Class
+                    | AttributeTargets.Struct
+                    | AttributeTargets.Interface
+                    | AttributeTargets.Property
+                    | AttributeTargets.Field, AllowMultiple = true)]
+    public class AdaptAsAttribute : Attribute
+    {
+        public AdaptDirectives AdaptDirective { get; }
+        public AdaptAsAttribute(AdaptDirectives directive)
+        {
+            this.AdaptDirective = directive;
+        }
+    }
+}
--- a/src/Mapster.Core/Attributes/AdaptAsAttribute.cs
+++ b/src/Mapster.Core/Attributes/AdaptAsAttribute.cs
@@ -7,10 +7,10 @@ namespace Mapster
                     | AttributeTargets.Interface
                     | AttributeTargets.Property
                     | AttributeTargets.Field, AllowMultiple = true)]
-    public class AdaptAsAttribute : Attribute
+    public class AdaptWithAttribute : Attribute
     {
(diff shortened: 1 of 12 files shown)


DocSvartz:
Explanations regarding tests marked Ignore
I don’t understand what exactly they wanted to check in them?
1) If it is possible to map non-public fields to the record constructor, then the Source should have had them
2) If you map fields from the Source to hidden fields in the record, then this should not work at all. The record is always activated through the constructor and they would have gotten there anyway. Because the sources are public.

andrerav:
Closed and reopened to re-run the workflow.

andrerav:
Thank you @DocSvartz!

PR #655, opened 2023-11-03
netstandard 2.0 compile

Jojoshua:


andrerav:
Thanks @Jojoshua, I will take a look. At first glance it seems the test suite failed, but I haven't taken a closer look yet. Did you run the tests locally?

DocSvartz:
@andrerav  For it to work, also need to remove all links to MapsterDynamic.
and install from nuget Microsoft.CSharp in MapsterCore

not Working map to interface 

and this tests 
TestMapFromSecondSourceObject()
TestIMapFrom_WhenMethodImplemented()
TestIMapFrom_WhenMethodIsNotImplemented()

DocSvartz:
I understand. This will work without global refactoring the code. 
With preprocessor directives.
But only when targeting NetFramework 4.8 or 4.7 possible.
i checked on local machine with 4.8 all build and test green 

IMapFrom dont working

Jojoshua:
> Thanks @Jojoshua, I will take a look. At first glance it seems the test suite failed, but I haven't taken a closer look yet. Did you run the tests locally?

I didn't run any tests locally. I only had 30 mins yesterday with a goal to make it build without errors. 

It went smooth except for the one method I had to comment out for imap because default method interfaces are not supported. I feel like this is could be easily changed when I understand what it's for. 

This was my first look at the codebase,  I will try to give this another 30 today and run the tests. If anyone can get tests working before I get to it feel free.

DocSvartz:
@Jojoshua  You get Mapster.dll  from netstandart 2.0 before compile ?

Jojoshua:
> @Jojoshua You get Mapster.dll for netstandart 2.0 after compile.
> 
> upd: i get dll for netstandart 2.0 I didn't see that it is necessary to install this
> 
> ```
> 
> <ItemGroup>
>     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
>     <PackageReference Include="System.Reflection.Emit" Version="4.7.0" />
>   </ItemGroup>
> ```

The project wouldn't compile for me without adding those. System.Reflection.Emit has netstandard 2.0 support but if you don't require it standalone it's going to want to pull it from net6/7.

Jojoshua:
@andrerav Ok, tests are now passing for me locally. Does the PR automatically run the tests or do you need to execute it?

src/Mapster/Interfaces/IMapFrom.cs
```diff
@@ -2,8 +2,12 @@ namespace Mapster;
 
 public interface IMapFrom<TSource>
 {
-    //public void ConfigureMapping(TypeAdapterConfig config)
-    //{
-    //    config.NewConfig(typeof(TSource), GetType());
-    //}
+
```
DocSvartz, line 5: This fix test, but not fix problem. 
Temporarily target mapster.Test to net48 and you see it.
IMapFrom  looks like for a netstandard 2.0  should be a different implementation or completely cut out.

Jojoshua: Ok, I will have to try that next. Do you have a suggestion for fixing it?

DocSvartz: I figured out how to cut it out. But there is no way to make it work yet.
https://github.com/DocSvartz/Mapster/tree/net-standart-2.0
Last two commits to Related to this issue.

Jojoshua: Is it necessary to have it target net48? When it previously targeted net6.0/netcor3.1 wasn't this all working in .net framework pre Mapster 7.4?

DocSvartz: Before version 7.4.0 Mapster was target at netstandard 2.0, 1.3. those. The development itself ensured that features that are available in newer TFM would not be explicitly used.

Compatibility is always backwards. Net 6.0 can test everything that came before it. Because he himself supports it.
Here is a different case.
To make it work you had to look for workarounds  you build in and disable features that are not included in the target platform.
Therefore, I think that it is necessary to test that the behavior of the api does not differ for different target platforms.

My answer is, no, not necessarily. But then there is no guarantee that everything will work as intended in an environment that really only supports netstandard 2.0.

DocSvartz:
@andrerav  How do you see Mapster's public api extension strategy? 
1) Support only available functionality to target platform
2) Support only functionality that will be on all target platforms

src/Mapster/Interfaces/IMapFrom.cs
DocSvartz: ImapFrom.cs
```
#if NETSTANDARD2_1_OR_GREATER
public interface IMapFrom<TSource>
{

    {
        config.NewConfig(typeof(TSource), GetType());
    }
    
}
#endif

#if !NETSTANDARD2_1_OR_GREATER
public interface IMapFrom<TSource>
{
    public void ConfigureMapping(TypeAdapterConfig config);
    
}
public static class MapFromLegacyHelper 
{
    public static void MapFromConfigureDefault<TSource>(this IMapFrom<TSource> mapFrom, TypeAdapterConfig config)
    {
        config.NewConfig(typeof(TSource), mapFrom.GetType());
    }

}

#endif
```

From Test where need implementation

```
#if !NETSTANDARD2_1_OR_GREATER
    {
        this.MapFromConfigureDefault(config);
    }
#endif
```

DocSvartz: It looks like it's just a registration helper.
In tests, even if you don’t specify an implementation, it will still work.
```

{
}
```

andrerav:
@DocSvartz @Jojoshua IMapFrom was added as a new feature in 7.4.0 (see #533). It's perfectly fine to leave this out from a netstandard2.0-compatible version of Mapster in my opinion. It's impossible for anyone to have put that interface to use in that user group :) We can add a note about this in the documentation.

Jojoshua:
> @DocSvartz @Jojoshua IMapFrom was added as a new feature in 7.4.0 (see #533). It's perfectly fine to leave this out from a netstandard2.0-compatible version of Mapster in my opinion. It's impossible for anyone to have put that interface to use in that user group :) We can add a note about this in the documentation.

That makes sense. With that in mind perhaps the PR is good to go? Could you try running the tests again?

andrerav:
@DocSvartz @Jojoshua The tests passed, so this looks promising :) I will check that the packaging works as expected before merging, but I don't expect any problems.

DocSvartz:
I am writing this more as a warning to the author that he could potentially get behavior that he did not expect. If he uses it where he wanted (an environment where there is access only to the netstandard 2.0 ). Because the tests do not use the target library for this case.

Issue #656, opened 2023-11-07
Cannot map properties that have differt names and init access modifier


nicoarm93:
Hi,
I have a problem with version 7.4. If I want to map two properties with different names and the destination has  a different name and the `init` modifier, Mapster throws a NullReferenceException. This happens when in the configuration of Mapster I set `PreserveReference` to `true`.

Here's the code

```internal class MapsterTest
{
    public Mapper Mapper { get; }

    public MapsterTest()
    {
        var config = new TypeAdapterConfig();
        config.Default.Settings.PreserveReference = true;
        config.Scan(Assembly.GetAssembly(typeof(MapsterTest)));
        config.Compile();
        Mapper = new Mapper(config);
    }
}

internal class MyClass
{
    public int MyProperty { get; set; }
}

internal class DtoClass
{
    public int MyPropertyDto { get; init; }
}
internal class Register : IRegister
{
    void IRegister.Register(TypeAdapterConfig config)
    {
        config.NewConfig<MyClass, DtoClass>()
            .Map(x => x.MyPropertyDto, x=> x.MyProperty);
    }
}
`````
And here's the mapping

```var mapster = new MapsterTest();

var from = new MyClass { MyProperty = 1};

var to = mapster.Mapper.Map<DtoClass>(from);
`````
(The problem does not manifest if the properties have the same name or another modifier (`set;` or `private set;`)
This code works fine with version 7.3

DocSvartz:
linked #422 

@nicoarm93  @andrerav   location of this bug found )

Mat3oo:
In my case similar regression (_Mapster throws a NullReferenceException_) between 7.4.0 and 7.3.0 is not related to `PreserveReference`. It's about `init` only properties, when mapping to target object.
Below code works fine with `Mapster 7.3.0`, but doesn't work with `Mapster 7.4.0`.

```c#
using Mapster;

TypeAdapterConfig<Core.RootSource.NestedSource, Core.RootDestination.NestedDestination>
    .NewConfig()
    .Map(dest => dest.NameDest, src => src.NameSrc);

var dest = new Core.RootDestination { MyProperty = null };

var mapped = Core.RootSource.Create().Adapt(dest);

Console.WriteLine(mapped);

namespace Core
{
    public class RootSource
    {
        public required NestedSource? MyProperty { get; set; }

        public static RootSource Create()
        {
            return new RootSource
            {
                MyProperty = new NestedSource
                {
[...]

Issue #422, opened 2022-04-04
Mapster.Tool - Error generating two-way mappings with init properties.


DonnelCyril:
Best to explain the issue with an eg:

Below is my domain model

```
public class SalesOrder
{
    public Guid Id { get; init; }
    public decimal Total { get; init; }
    public string Status { get; init; }
}
```

and I am trying to create a corresponding Dto using the Mapster.Tool

```
    <Target Name="Mapster">
        <Exec WorkingDirectory="$(ProjectDir)" Command="dotnet mapster model -a &quot;$(TargetDir)$(ProjectName).dll&quot; -n &quot;Mapster.TestBench.Dtos&quot; -o &quot;Dtos/Generated&quot;" />
        <Exec WorkingDirectory="$(ProjectDir)" Command="dotnet mapster extension -a &quot;$(TargetDir)$(ProjectName).dll&quot; -n &quot;Mapster.TestBench.DtoMapper&quot; -o &quot;Dtos/Generated&quot;" />
    </Target>

```

The Dto object gets generated as expected.

```
    public partial class SalesOrderDto
    {
        public Guid Id { get; set; }
        public decimal Total { get; set; }
        public string Status { get; set; }
    }
```

But the mapper code generated doesn't compile.

  ```
      public static SalesOrder AdaptTo(this SalesOrderDto p2, SalesOrder p3)
        {
            if (p2 == null)
            {
                return null;
            }
            SalesOrder result = p3 ?? new SalesOrder();
            
             //===========================================

                // Id, Total & Status are init properties in the SaleOrder type and fails to compile.
    
                result.Id = p2.Id;
                result.Total = p2.Total;
                result.Status = p2.Status;

             //===========================================

            return result;
            
        }
```

It is in this second variation of _Adapt_ method that takes in both an Dto & Entity object, and tries to update the Entity object with the values from Dto object.

The other method generated works as expected.
```
public static SalesOrder AdaptToSalesOrder(this SalesOrderDto p1)
{
    return p1 == null ? null : new SalesOrder()
    {
        Id = p1.Id,
        Total = p1.Total,
        Status = p1.Status
    };
}

```

If this is a bigger piece of work, even a flag to avoid creating the second method will be an interim fix.

My CodeGenerationRegister as below if it helps.

```
public class ModelGenerationRegister : ICodeGenerationRegister
{
    public void Register(CodeGenerationConfig config)
    {

        config.AdaptTwoWays($"[name]Dto")
[...]

PR #658, opened 2023-11-12
Fix Issue #656. Mapping init Property

DocSvartz:
Fix Issue #656 

Partial fix  issue #422 
   Fix MapToTarget generation to not Public setters (init Prorperty also working in Mapster 7.4.0)
Not Fix AdaptTo - need use `.MapToConstructor(true)`;

example

```
public class SalesOrder
{
   public SalesOrder(){}
   public SalesOrder(Guid id, decimal total, string status)
    {
        Id = id;
        Total = total;
        Status = status;
    }

    public Guid Id { get; protected set; }
    public decimal Total { get; protected set; }
    public string Status { get; protected set; }
}

Bad generation :

public static SalesOrder AdaptToSalesOrder(this SalesOrderDto p1)
{
    return p1 == null ? null : new SalesOrder()
    {
        Id = p1.Id, 
        Total = p1.Total,
        Status = p1.Status
    };
}
```

when using `.MapToConstructor(true)`

 ```
Good generation :

public static SalesOrder AdaptToSalesOrder(this SalesOrderDto p1)
{
    return p1 == null ? null : new SalesOrder(p1.Id, p1.Total, p1.Status) {};
}



```

### DIFF
--- a/src/Mapster.Tests/WhenMappingInitPropertyRegression.cs
+++ b/src/Mapster.Tests/WhenMappingInitPropertyRegression.cs
@@ -14,29 +14,29 @@ namespace Mapster.Tests
         [TestMethod]
         public void InitPropertyWorked()
         {
-            MapsterTest mapster = new MapsterTest();
+            MapsterTest656 mapster = new MapsterTest656();
 
             var from = new MyClass656 { MyProperty = 1 };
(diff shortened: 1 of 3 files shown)


andrerav:
Fixed by #745.

PR #669, opened 2023-12-15
fix #656

### LINKED ISSUE #656: Cannot map properties that have differt names and init access modifier

SimTsai:
fix issue, when Dst init property name miss match Src property name.

### DIFF
--- a/src/Mapster/Adapters/ClassAdapter.cs
+++ b/src/Mapster/Adapters/ClassAdapter.cs
@@ -134,7 +134,7 @@ namespace Mapster.Adapters
                         {
                             var destinationPropertyInfo = (PropertyInfo)member.DestinationMember.Info!;
                             adapt = destinationPropertyInfo.IsInitOnly()
-                                ? SetValueByReflection(destination, (MemberExpression)adapt, arg.DestinationType)
+                                ? SetValueByReflection(member, (MemberExpression)adapt)
                                 : member.DestinationMember.SetExpression(destination, adapt);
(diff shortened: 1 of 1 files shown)


andrerav:
Please use the development branch for pull requests.

Issue #714, opened 2024-06-14
System.IO.FileLoadException: Could not load file or assembly ... Could not find or load a specific file with .NET 9 preview package


boylec:
This is a similar issue to #635 in that an assembly is not being loaded when running Mapster.Tool 8.4.1-pre-1, but the error message is slightly different.

In #635 there is a FileLoadException with "The system cannot find the file specified."

I was seeing this with a different assembly and fixed it using the recommendations from that issue.

Fast forward ->

Then I added System.Text.Json for .NET 9 preview (which targets .NET 8)

Now I'm getting a System.IO.FileLoadException with "Could not load file or assembly ... Could not find or load a specific file"

.csproj:
```
<PropertyGroup>
   ...
   <TargetFramework>net8.0</TargetFramework>
   <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
   ...
</PropertyGroup>
...
<PackageReference Include="System.Text.Json" Version="9.0.0-preview.5.24306.7" />
<PackageReference Include="Mapster" Version="7.4.1-pre01" />
...

  <ItemGroup>
    <Generated Include="$(ProjectDir)\**\*.g.cs" />
    <GeneratedMappersFolder Include="_mappers" />
  </ItemGroup>

  <Target Name="Mapster" AfterTargets="AfterBuild">
    <Exec WorkingDirectory="$(ProjectDir)" Command="dotnet tool restore" />
    <Exec WorkingDirectory="$(ProjectDir)" Command="dotnet mapster extension -a &quot;$(TargetDir)$(ProjectName).dll&quot;" />
    <Exec WorkingDirectory="$(ProjectDir)" Command="dotnet mapster mapper -a &quot;$(TargetDir)$(ProjectName).dll&quot; -o @(GeneratedMappersFolder)" />
    <Message Text="GeneratedMappersFolder: @(GeneratedMappersFolder)" />
  </Target>

  <Target Name="CleanGenerated" BeforeTargets="AfterBuild">
    <Delete Files="@(Generated)">
      <Output TaskParameter="DeletedFiles" ItemName="FilesDeleted" />
    </Delete>
    <Message Text="Files deleted: @(Generated)" />
  </Target>
```

dotnet-tools.json
```
  ...
    "mapster.tool": {
      "version": "8.4.1-pre01",
      "commands": ["dotnet-mapster"],
      "rollForward": false
    }
   ...
}
```

stack trace:
```
[...]

PR #715, opened 2024-06-14
Use assembly loading context to isolate loaded assemblies from AppDomain #714

boylec:
See issue #714 

I was running into issues running Mapster.Tool against projects with .NET 9 preview assemblies references (either directly or transitively from other dependencies.)

After sleuthing through dotnet/runtime CLR code I ran into this issue which seemed pertinent:

https://github.com/dotnet/runtime/discussions/102981

This PR is based off of recommendations by @jkotas there about how to load assemblies in an isolated way. It appears to have solved my issue with Mapster.Tool 8.4.1-pre01 mentioned.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -181,3 +181,8 @@ project.lock.json
 /src/.vs
 /.vs
 src/.idea
+
+# VS Code settings
+.vscode/launch.json
+.vscode/settings.json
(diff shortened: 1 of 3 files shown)


boylec:
need to work on this a bit more

PR #716, opened 2024-06-15
Use assembly loading context to isolate loaded assemblies from AppDomain #714

boylec:
# Mapster.Tool update to solve issue #714

## TL; DR

Replace Assembly.LoadFrom calls with something more robust that prevents assembly/type collisions between the assemblies in the running Mapster.Tool process (such as .NET framework assemblies) and the scanned assemblies - while also preserving the ability to scan for attributes that are in the Mapster assembly (using share Assembly resolution).

See [Pertinent MSFT Docs](https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/understanding-assemblyloadcontext#shared-dependencies)

## More detail
Added an AssemblyLoadContext implementation for dynamically loading assemblies when scanning in an isolated way so that they don't have collisions with all of the assemblies loaded by Mapster by default (for example framework assemblies such as System.Text.Json).

The implementation also allows the setting of a subset of assembly names that will be shared with AssemblyLoadContext.Default (the default context that all assemblies are resolved from if no custom ALC were created). Some scanned types need to be owned by the default assembly context such as those from the "Mapster" assembly.

The end result is that we can resolve types from the target assembly being scanned, and we can do things like .GetCustomAttribute<MappingAttribute>() and the attribute will be found because it comes from a shared assembly across the default (runtime) AssemblyLoadContext and the custom one used to avoid collisions.

### DIFF
--- a/src/nuget.config
+++ b/src/nuget.config
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
+    <clear />
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+    <add key="truevelocity" value="https://pkgs.dev.azure.com/truevelocity/_packaging/truevelocity/nuget/v3/index.json" />
+  </packageSources>
+</configuration>
--- a/src/nuget.config
+++ b/src/nuget.config
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<configuration>
-  <packageSources>
-    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
-    <clear />
-    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
-    <add key="truevelocity" value="https://pkgs.dev.azure.com/truevelocity/_packaging/truevelocity/nuget/v3/index.json" />
-  </packageSources>
-</configuration>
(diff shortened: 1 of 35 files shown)


boylec:
@andrerav what are the chances this can get merged any time soon?

Our org relies on this package and this is a blocker for us.

We can run off of our fork if it's going to be awhile.

Thanks!

src/nuget.config
andrerav, line 1: Was it intentional to include this file?

andrerav (changes requested):

src/nuget.config
andrerav: @boylec Happy new year :) Please see my above comment.

boylec: no this was not intentional - good catch

andrerav: Everything seems good otherwise. If you can nuke that file and resolve the conflict I'll get this merged and packaged up with the .NET 9 update.

boylec:
@andrerav done and done thanks!

andrerav (approved):

andrerav:
Thank you for this awesome contribution @boylec. If you have some extra time to spare, please update the [Wiki](https://github.com/MapsterMapper/Mapster/wiki) to reflect the new API changes.

boylec:
@andrerav as far as I can tell there are no changes to consumer-facing APIs of the library.

It has been a long time since I made this change though, are you seeing something specific?

andrerav:
Ah, disregard that. I had a brief lapse of reason :)

Issue #723, opened 2024-08-06
.Ignore() for interface pairs throws ArgumentException


luczito:
Suppose we have two interfaces which are identical except that one has a field the other does not. 

```C#
public interface Interface1
{
    public string Inter { get; set; }
    
    public string Ignore { get; }
}
```

```C#
public interface Interface2
{
    public string Inter { get; set; }
}
```

When creating a mapping config we would simply ignore this field as:

```C#
config.NewConfig<Interface1, Interface2>() 
   .TwoWays()
   .Ignore(dest => dest.Ignore);
``` 

However this throws the error:
"System.ArgumentException : Incorrect number of arguments for constructor"
  
The only fix as far as i can tell is to explicitly map this property to something E.G:

```C#
string temp = null;
config.NewConfig<Interface1, Interface2>()
   .TwoWays()
   .Map(dest => dest.Ignore, src => temp);
```

Or to hardcode what class to use when this mapping pair is hit:

```C#
config.NewConfig<Interface1, Interface2>()
   .Twoways()
   .ConstructUsing(src => new DerivedClass());
```

Both of these fixes are suboptimal, are there any other way to get around this issue?

DocSvartz:
Hello @luczito I think I found the place where this error occurs.
But are you sure about the specified configuration or the description of the interfaces provided?
On this basis, I can set the condition you specified only in the reverse configuration. 

```
  TypeAdapterConfig<Interface2, Interface1>
      .NewConfig()
      .TwoWays()
      .Ignore(dest => dest.Ignore);
```

(And it doesn't make much sense. Since the value won't be transferred anyway. If this field is not in the source).

PR #754, opened 2025-01-07
Fix issue #723 - Add Support ignore member when creating a constructor in DynamicTypeGenerator

DocSvartz:


DocSvartz:
Hello @andrerav .

If the interface contains at least one property with a non-public setter, then all properties will be initialized via the constructor.
Custom logic for ignoring them is not supported.

I found a way to add this support, but maybe it was not intended that way?

src/Mapster/Utils/DynamicTypeGenerator.cs
```diff
@@ -175,5 +186,27 @@ private static void CreateMethod(TypeBuilder builder, MethodInfo interfaceMethod
 
             builder.DefineMethodOverride(classMethod, interfaceMethod);
         }
+
+        private static List<FieldBuilder> DropIgnorMemebers(List<FieldBuilder> fields)
```
lofcz, line 190: possibly a typo: `DropIgnoredMembers`

src/Mapster/Utils/DynamicTypeGenerator.cs
```diff
@@ -22,6 +22,16 @@ internal static class DynamicTypeGenerator
         private static readonly ConcurrentDictionary<Type, Type> _generated = new ConcurrentDictionary<Type, Type>();
         private static int _generatedCounter;
 
+        private static IgnoreDictionary ignoreMembers;
```
lofcz, line 25: shouldn't this be nullable?

src/Mapster/Utils/DynamicTypeGenerator.cs
DocSvartz: yes of course. This is not the final solution, you could say MVP :)

Issue #755, opened 2025-01-07
Regression in mapping interfaces, introduced in #649


lofcz:
Merging of #649 introduced regressions in the mapping of interfaces. A minimal test:

```cs
[TestMethod]
public void MappingToInterface_Broken()
{
    SampleInterfaceCls source = new SampleInterfaceCls
    {
        ActivityData = new SampleActivityData
        {
            Data = new SampleActivityParsedData
            {
                Steps = new List<string> { "A", "B", "C" }
            }
        }
    };

    SampleInterfaceCls dest = source.Adapt<SampleInterfaceCls>(); // works pre #649, crashes after
}

public class SampleInterfaceCls
{
    [Newtonsoft.Json.JsonIgnore]
    public IActivityData? ActivityData { get; set; }

    public SampleInterfaceCls()
    {

    }

    public SampleInterfaceCls(IActivityData data)
    {
        SetActivityData(data);
    }

    public void SetActivityData(IActivityData data)
    {
        ActivityData = data;
    }
}

public class SampleActivityData : IActivityData
{
    public SampleActivityParsedData Data { get; set; }
}

public class SampleActivityParsedData
{
    public List<string> Steps { get; set; } = new List<string>();
}
```

DocSvartz:
Hello @lofcz As far as I understand, my test from the [issue](https://github.com/MapsterMapper/Mapster/pull/756#issuecomment-2577286411) does not work on 7.4.0. ?

Is this behavior really necessary?   _It turned out to be more complicated than I thought at first :)_

lofcz:
I believe it's a problem of expectations vs what Mapster _does_. For this specific scenario, I've ended up writing a new library https://github.com/lofcz/FastCloner which does deep cloning _primarily_ rather than mapping (which I used for deep cloning). I still believe #756 should be merged. Your input would be appreciated.

andrerav:
Fixed by #756

andrerav:
@lofcz @DocSvartz I've pushed a new set of prerelease packages now (pre02).

PR #758, opened 2025-01-11
 Fix issue #755 - Revert to behavior from 7.4.0 and new features

DocSvartz:
New:

1. Mapping of the interface itself (IActive->IActive) not generate a new implementation from DynamicTypeGenerator.
2. Mapping Derived Interface to Base Interface does not generate a new implementation from DynamicTypeGenerator

### DIFF
--- a/src/Mapster.Tests/WhenMappingRecordRegression.cs
+++ b/src/Mapster.Tests/WhenMappingRecordRegression.cs
@@ -278,6 +278,83 @@ namespace Mapster.Tests
         }
 
 
+        [TestMethod]
+        public void MappingInterfaceToInterface()
+        {
+            TypeAdapterConfig<IActivityData, IActivityDataExtentions>
+                .ForType()
+                .Map(dest => dest.TempLength, src => src.Temp.Length);
+
+
+            var sourceBase = new SampleInterfaceClsBase
+            {
+                ActivityData = new SampleActivityData
+                {
+                    Data = new SampleActivityParsedData
+                    {
+                        Steps = new List<string> { "A", "B", "C" }
+                    },
+                    Temp = "Temp data"
+
+                }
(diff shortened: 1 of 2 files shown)


DocSvartz:
@andrerav  @lofcz 

This completely emulates the logic of the Primitive adapter, which would be passed in 7.4.0 this case ( map IActivity -> IActivity).
The only question is caused by this case Base.Adapt(Derived).
But as far as I understand in 7.4.0 this also would not work. - Property Data does not receive state from Source

src/Mapster.Tests/WhenMappingRecordRegression.cs
```diff
@@ -278,6 +278,83 @@ public void MultiCtorAndInlineRecordWorked()
...
+            };
+
+            var TargetBase = sourceBase.Adapt<SampleInterfaceClsBase>();
+            var targetDerived = sourceDerived.Adapt<SampleInterfaceClsDerived>();
+            var update = targetDerived.Adapt(TargetBase);
+
+            var targetExtention = sourceExt.Adapt<SampleInterfaceClsExtentions>();
+
+
+            var updExt = targetDerived.Adapt(targetExtention);
```
DocSvartz, line 338: Base.Adapt(Derived) case

andrerav:
@DocSvartz Please synchronize your branch and I can do a review. I'll release the current code in development as pre02, and we can release this as pre03. Sounds good?

DocSvartz:
@andrerav  OK

DocSvartz:
@andrerav I resolved the merge conflict. 
Let me know if these changes are acceptable.
I would like to continue working on #723 with them in mind :)

andrerav:
Thank you @DocSvartz!

PR #759, opened 2025-01-13
Fix null map to target regression

DocSvartz:


andrerav:
Thank you @DocSvartz!

Issue #456, opened 2022-07-22
Mapping fails when a record property is `Ignore`d


amantinband:
The following will throw an exception:

```csharp
public record User(string Name);
```

```csharp
public record UserDto(string Name);
```

```csharp
TypeAdapterConfig<User, UserDto>.NewConfig()
    .Ignore(dest => dest.Name);

var user = new User("Amichai");
var userDto = user.Adapt<UserDto>(); // <-- throws Mapster.CompileException: Error while compiling, with inner exception: System.ArgumentException: Incorrect number of arguments for constructor
```

Is this by design? Didn't see this documented

skuirrels:
This needs to be solved. It precludes the use of Mapster.

stagep:
In order to identify issues that are still active, we are closing issues that we believe are either resolved or are dormant. If your issue is still active then please reopen. Thanks.

DocSvartz:
@amantinband @andrerav  Yes, this not supporting in origin design. 

I hope add this :)

DocSvartz:
linked #707 
@andrerav  ignoring works too well :)
To create a construct, data about the presence of such a variable is not received at all.
fix is almost complete :)

DocSvartz:
linked #458

Issue #707, opened 2024-05-14
Mapping with constructor and default values for ignored properties?


DumboJetEngine:
I try to map from class A to class B, where B is missing a property of A.
Usually, to do this you have to use `.Ignore(e => e.Property)` but I use a non-default constructor for initialization of the destination class (because I need to validate the incoming data within the constructor) and this fails.

Apparently, Mapster removes the ignored parameter from the constructor invocation parameters instead of using the default parameter value. And this inevitably fails.

Here is some sample code:

```
using Mapster;
using System.Reflection;

var config = new TypeAdapterConfig()
{
    RequireExplicitMapping = false,
    RequireDestinationMemberSource = true,
};

config.Default
    .PreserveReference(true)
    .MapToConstructor(true)
    .Unflattening(false)
    .IgnoreNullValues(false)
    .NameMatchingStrategy(new NameMatchingStrategy
    {
        SourceMemberNameConverter = input => input.ToLowerInvariant(),
        DestinationMemberNameConverter = input => input.ToLowerInvariant(),
    })
    .ShallowCopyForSameType(false);

var ctorToUse = GetConstructor<B>();
config
    .NewConfig<A, B>()
    .Ignore(e => e.Id) // <--------------- The code fails with or without this invocation.
    .MapToConstructor(ctorToUse);

var mapper = new MapsterMapper.Mapper(config);

static ConstructorInfo? GetConstructor<TDestination>()
{
    var parameterlessCtorInfo = typeof(TDestination).GetConstructor(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public, new Type[0]);

    var ctors = typeof(TDestination).GetConstructors(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public);
    var validCandidateCtors = ctors.Except(new[] { parameterlessCtorInfo }).ToArray();
    var ctorToUse = validCandidateCtors.Length == 1
        ? validCandidateCtors.First()
        : validCandidateCtors.OrderByDescending(c => c.GetParameters().Length).First();

    return ctorToUse;
}


var a = new A
{
    Text = "test",
};

var docKind = mapper.Map<B>(a); // <------ This fails.

class A
{
    public string? Text { get; set; }
}

class B
{
    public int Id { get; private set; }
    public string Text { get; private set; }

    public B(int id, string text)
    {
        Id = id;
        Text = text;
    }
}
```

[...]

Issue #458, opened 2022-07-25
Incorrect mapping of an ignored property


PopMal:
### Mapster Setting:
`TypeAdapterConfig.GlobalSettings.Compiler = exp => exp.CompileWithDebugInfo();`
`TypeAdapterConfig.GlobalSettings.Default.PreserveReference(true);
 TypeAdapterConfig.GlobalSettings.RequireDestinationMemberSource = true;
 TypeAdapterConfig.GlobalSettings.Scan(typeof(MappingConfig).Assembly);`
`TypeAdapterConfig.GlobalSettings.Compile();`

### Entity Configuration:
`TypeAdapterConfig<Entity, Dto>.NewConfig()
                .Map(entity=> entity.Id, dto => dto.Id)
                .Ignore(entity=> entity.PropToIgnore)`

### My entity (PropToIgnore):

public class PropToIgnore    {
         private LexBotInfo() { }
         public LexBotInfo(string name, string desc)
        {
            Desc = desc;
            Name = name;
        }
        public string Name { get; private set; }
        public string Desc{ get; private set; }
}

### Runtime error:

[...]

PR #763, opened 2025-01-21
Fix Issue #456 #707 #723 - Ignore() processing improvements

DocSvartz:
#456  When explicitly set to ignore():

1. Instance of a class can be created with default values ​​for parameters.
2. Explicitly set to ignore()  not mark param as not matched when config `RequireDestinationMemberSource = true,`

#707 #723  RecordTypes and Generated Type to Interface received support Ignore() from Ctor parameters

### DIFF
--- a/src/Mapster.Tests/WhenIgnoreMapping.cs
+++ b/src/Mapster.Tests/WhenIgnoreMapping.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Linq;
+using System.Reflection;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Newtonsoft.Json;
 using Shouldly;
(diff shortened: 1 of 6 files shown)


DocSvartz:
@andrerav  @lofcz  This will also resolve issue #723. 
By calling the constructor with a default parameter. 
Instead of creating a constructor without this parameter. As in the previous solution #761.

Which of the two solutions presented is better?

I think seems this solution is better.  Especially if these created classes are used in other mappings. 🤔

andrerav:
Great @DocSvartz. Whenever you have PR's that you think is ready to be merged, could you please assign me as reviewer and I will review as soon as I can?

DocSvartz:
Ok

DocSvartz:
linked #458 mapping ctor error

DocSvartz:
@andrerav This not ready for RecordType (
In terms of the fixes. It's more convenient for me to add them as one PR.
But then it will be more difficult to review.

Or I can enable it only for Classes.
And for records this will already be included as part of the fix MapTotarget behavior for RecordType (new PR)?

DocSvartz:
I think this can be closed. I improved it and added to PR #769.
In which I hope I finally managed to fix MapToTarget for RecordType ))

Issue #413, opened 2022-03-21
IgnoreNonMapped is ignored when using RequireDestinationMemberSource


cytoph:
I have the following config:

```
var config = new TypeAdapterConfig()
{
    // other configuration here
    RequireDestinationMemberSource = true,
};

// other mappings here

config.NewConfig<DataRow, Document>()
    .Map(d => d.Id, dr => dr.Field<int>("Id"))
    .IgnoreNonMapped(true);
```

When I now use `config.Compile()` to check the configurations I get an exception that on mapping DataRow to Document all properties (except Id of course) where neither configured to be mapped or ignored. My workaround for now is to use `.RequireDestinationMemberSource(false)` just for this one mapping, but shouldn't `IgnoreNonMapped` still work as inteded when using `RequireDestinationMemberSource = true`?

polom01:
@andrerav

floatas:
@andrerav does this project participate in hacktoberfest ? There are some issues that the community could help out with.

duhapunk:
UP

ongun23:
That is still happening.

DocSvartz:
@andrerav  I think it is related to setting contradictory instructions) 

This setting needs to be renamed. If that's what it's supposed to be. 

.IgnoreNonMapped(true)  != Ignore(all property) 

This means don't use the default mapping. -> Only [Manual Mapping](https://github.com/MapsterMapper/Mapster/blob/04ac871b55828c3909b6cee4764e6fab40db3983/src/Mapster/Adapters/BaseClassAdapter.cs#L37) mode.

All properties remain in the mapping processing -> not map by default, not manual map -> Exception

DocSvartz:
@andrerav  I open PR #765  from with the solution to this issue )

DocSvartz:
This should be fixed in new release versions.
If your issue is still active, please reopen it or create new issue. Thanks.

PR #765, opened 2025-01-23
Fix issue #413 - add Skip check RequireDestinationMemberSource when use IgnoreNonMapped(true)

DocSvartz:
Fix issue #413


DocSvartz:
I try convert to real Ignore from not mapping members

PR #766, opened 2025-01-23
FIx Issue #537 - Step 4  - Fix MapToTarget behavior From RecordType

DocSvartz:
before  from 
```
record TestRecord()
{
    public int X { set; get; }
}
record TestRecordY() 
{
    public int X { set; get; }
    public int Y { set; get; }
}

var source = new TestRecord() { X= 200};
var dest = new TestRecordY() {X = 100, Y= 500}

```

```
var adapt = source.Adapt(dest)  // always adapt == { X=200, Y=0}
```

after 

```
var adapt = source.Adapt(dest)  // adapt == { X=200, Y=500}  Correct MapTotarget
```

equal behavior when modified using with 

```
var adapt = dest with {X=source.X};
```

### DIFF
--- a/src/Mapster/Adapters/BaseClassAdapter.cs
+++ b/src/Mapster/Adapters/BaseClassAdapter.cs
@@ -54,6 +54,15 @@ namespace Mapster.Adapters
                     Destination = (ParameterExpression?)destination,
                     UseDestinationValue = arg.MapType != MapType.Projection && destinationMember.UseDestinationValue(arg),
                 };
+                if (arg.MapType == MapType.MapToTarget && getter == null && arg.DestinationType.IsRecordType())
+                {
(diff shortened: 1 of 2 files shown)


DocSvartz:
@andrerav  working From Ctor ignoring I found that MapToTarget was not working for RecordType. 
Looks like I missed that in my original solution (
I didn't add test to avoid creating a merge conflict. Although most likely they will be anyway, Sorry

DocSvartz:
@andrerav It also looks like this case is also not working correctly from Record (When using Ignore())

```
var adapt = dest with { };
```

I will add a separate PR after the final decision on the possibility of calling Ctor with default value. 
I still need to figure out how to combine it. )

DocSvartz:
@andrerav  This not ready for RecordType (
In terms of the fixes. It's more convenient for me to add them as one PR.
But then it will be more difficult to review.

Or I can enable it only for Classes. 
And for records this will already be included as part of the fix  MapTotarget behavior for RecordType  (new PR)?

PR #769, opened 2025-01-25
Improvements to Ignore() processing and Fix MapToTarget behavior for RecordTypes 

DocSvartz:
Fix Issue #456 #707 #723

#456 When explicitly set to ignore():

Instance of a class can be created with default values ​​for parameters.
Explicitly set to ignore() not mark param as not matched when config RequireDestinationMemberSource = true,

#707  #723 RecordTypes and Generated Type to Interface received support Ignore() from Ctor parameters

Fix MapToTarget behavior for RecordTypes (Include using Ignore() )

before  from 
```
record TestRecord()
{
    public int X { set; get; }
}
record TestRecordY() 
{
    public int X { set; get; }
    public int Y { set; get; }
}

var source = new TestRecord() { X= 200};
var dest = new TestRecordY() {X = 100, Y= 500}

```

```
var adapt = source.Adapt(dest)  // always adapt == { X=200, Y=0}
```

after 

```
var adapt = source.Adapt(dest)  // adapt == { X=200, Y=500}  Correct MapTotarget
```

equal behavior when modified using with 

```
var adapt = dest with {X=source.X};
```
add Support from RecordType

- IgnoreNullValues()
- UseDestinationValue
- UseDestination value from Primitive type (and RecordType) AutoProperty

### DIFF
--- a/src/Mapster/Adapters/BaseClassAdapter.cs
+++ b/src/Mapster/Adapters/BaseClassAdapter.cs
@@ -1,10 +1,10 @@
-﻿using System;
+﻿using Mapster.Models;
+using Mapster.Utils;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
-using Mapster.Models;
-using Mapster.Utils;
 
 namespace Mapster.Adapters
 {
@@ -15,7 +15,7 @@ namespace Mapster.Adapters
 
         #region Build the Adapter Model
 
-        protected ClassMapping CreateClassConverter(Expression source, ClassModel classModel, CompileArgument arg, Expression? destination = null, bool ctorMapping = false)
+        protected ClassMapping CreateClassConverter(Expression source, ClassModel classModel, CompileArgument arg, Expression? destination = null, bool ctorMapping = false, ClassModel recordRestorMemberModel = null)
         {
             var destinationMembers = classModel.Members;
             var unmappedDestinationMembers = new List<string>();
@@ -56,12 +56,7 @@ namespace Mapster.Adapters
                 };
                 if (arg.MapType == MapType.MapToTarget && getter == null && arg.DestinationType.IsRecordType())
                 {
-                    var find = arg.DestinationType.GetFieldsAndProperties(arg.Settings.EnableNonPublicMembers.GetValueOrDefault()).ToArray()
-                                .Where(x => x.Name == destinationMember.Name).FirstOrDefault();
-
-                    if (find != null && destination != null)
-                        getter = Expression.MakeMemberAccess(destination, (MemberInfo)find.Info);
-
+                    getter = TryRestoreRecordMember(destinationMember, recordRestorMemberModel, destination) ?? getter;
                 }
                 if (getter != null)
                 {
@@ -138,7 +133,7 @@ namespace Mapster.Adapters
                    && ignore.Condition == null;
         }
 
-        protected Expression CreateInstantiationExpression(Expression source, ClassMapping classConverter, CompileArgument arg, Expression? destination)
+        protected Expression CreateInstantiationExpression(Expression source, ClassMapping classConverter, CompileArgument arg, Expression? destination, ClassModel recordRestorParamModel = null)
         {
             var members = classConverter.Members;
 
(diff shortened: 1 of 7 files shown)


src/Mapster/Adapters/BaseClassAdapter.cs
```diff
@@ -54,6 +54,15 @@ select fn(src, destinationMember, arg))
                     Destination = (ParameterExpression?)destination,
                     UseDestinationValue = arg.MapType != MapType.Projection && destinationMember.UseDestinationValue(arg),
                 };
+                if (arg.MapType == MapType.MapToTarget && getter == null && arg.DestinationType.IsRecordType())
+                {
+                    var find = arg.DestinationType.GetFieldsAndProperties(arg.Settings.EnableNonPublicMembers.GetValueOrDefault()).ToArray()
+                                .Where(x => x.Name == destinationMember.Name).FirstOrDefault();
+
```
DocSvartz, line 61: This can be simplified. on refactoring. Field information is already in the processing pipeline

src/Mapster/Adapters/BaseClassAdapter.cs
```diff
@@ -156,6 +175,20 @@ protected Expression CreateInstantiationExpression(Expression source, ClassMappi
...
                         getter = Expression.Condition(condition, getter, defaultConst);
                     }
+                    else
+                        if (arg.Settings.Ignore.Any(x => x.Key == member.DestinationMember.Name))
+                    {
+                        getter = defaultConst;
+
+                        if (arg.MapType == MapType.MapToTarget && arg.DestinationType.IsRecordType())
+                        {
+                            var find = arg.DestinationType.GetFieldsAndProperties(arg.Settings.EnableNonPublicMembers.GetValueOrDefault()).ToArray()
```
DocSvartz, line 185: Here it is also possible, but you need to get a list of Destination fields. Not constructor parameters.

src/Mapster/Adapters/RecordTypeAdapter.cs
```diff
@@ -56,9 +57,11 @@ protected override Expression CreateInlineExpression(Expression source, CompileA
             var exp = installExpr;
             var memberInit = exp as MemberInitExpression;
             var newInstance = memberInit?.NewExpression ?? (NewExpression)exp;
-            var contructorMembers = newInstance.Arguments.OfType<MemberExpression>().Select(me => me.Member).ToArray();
+            var contructorMembers = arg.DestinationType.GetConstructors()
+                    .OrderByDescending(it => it.GetParameters().Length).FirstOrDefault()
+                    .GetParameters().ToList();
```
DocSvartz, line 62: This can be simplified. information is already transferred to installExpr

src/Mapster/Adapters/RecordTypeAdapter.cs
```diff
@@ -93,8 +96,39 @@ protected override Expression CreateInlineExpression(Expression source, CompileA
...
+            if(arg.MapType == MapType.MapToTarget)
+                lines.AddRange(RecordIngnoredWithoutConditonRestore(destination, arg, contructorMembers));
+
             return Expression.MemberInit(newInstance, lines);
         }
+
+        private List<MemberBinding> RecordIngnoredWithoutConditonRestore(Expression? destination, CompileArgument arg, List<ParameterInfo> contructorMembers)
+        {
+           var members =  arg.DestinationType
+                            .GetFieldsAndProperties(arg.Settings.EnableNonPublicMembers.GetValueOrDefault())
```
DocSvartz, line 108: It is not possible to obtain such a data configuration in the current processing pipeline.
But it is possible to simplify it by obtaining it from a modified version of the install expression in the current solution.

DocSvartz:
@andrerav I left the comments for myself. So I wouldn't forget )
Most likely, we also need to add a Null check for the entire Destination object.
But I don't understand where yet ))

DocSvartz:
@andrerav  It's definitely ready for review )
Need fix UseDestinationValue for RecordType.  I'll try to come up with something in a separate PR. ))

DocSvartz:
@andrerav  It turned out to be much more difficult than I thought. 
But it looks like I managed to add almost all features from Classes to Record  :)

aos-p001:
Can this same functionality be applied/achieved in default config?

DocSvartz:
Hello Brad Jones,
I received a notification, but for some reason I don't see the place where
you left it on GitHub.
Now it is possible to ignore a property that participates in the
constructor only for classes in the case MapToTarget -> a.Adapt(b).
For records, this is not possible at all, since they are created through
the constructor.


пт, 7 февр. 2025 г. в 03:47, Brad Jones ***@***.***>:

PR #777, opened 2025-02-25
Feature: Allow for mapping properties containing periods

JMolenkamp:
The `InvokerModel` stores the property path as a period joined string. When building the expression for a property path, this string is split on the periods, which makes it impossible to have a property name that contains periods.

Obviously, this is kinda obscure, but I actually ran into this issue. It is possible to have a property name with a period when dynamically building types where pretty much all naming restrictions go out of the window, demonstrated in [this test](https://github.com/JMolenkamp/Mapster/blob/8828c76ff5acc6eb9f144b4f1a6c9826edea2afc/src/Mapster.Tests/WhenMappingPropertyNameContainingPeriod.cs#L14C1-L23C6).

With this PR, instead of storing a period joined string, the path is stored as a string array. This allows for storing a property name with a period as it is not split on usage. The tests still succeed, except for [`No_Errors_Thrown_With_Default_Configuration_On_Unmapped_Primitive`](https://github.com/MapsterMapper/Mapster/blob/155a72371a3119bf5fbf6d366952cc5c2c80e150/src/Mapster.Tests/WhenHandlingUnmappedMembers.cs#L18C1-L31C10), which failed to begin with.

If this change is considered, a skeptic review is required, even though the tests succeed. For now, this is just my initial attempt.

For some background: 
Our applications frequently have to communicate with industrial devices, for which the OPC UA protocol is often used. When reading or writing some structured data, an application type requires mapping to some node hierarchy provided by the OPC UA server. The configuration for this mapping is external, such that changes in property names/types or even structure do not require a rebuild of the application. Based on the configuration, types are built dynamically and Mapster is configured to map between the application type and the dynamic type. I am not really up to date with the naming restrictions on OPC UA nodes, but usage of periods is allowed, at least in some places.


DocSvartz:
Hello @JMolenkamp Excellent work!

1) Your test only checks that the configuration compiles successfully. If possible, add tests that check that the data is transfered successfully. (_sourse.Adapt<TDestination>() and _sourse.Adapt(_destination));

2) Have you checked the functionality of your solution when these Properties are used as constructor parameters?

JMolenkamp:
Hi @DocSvartz,

1. Compilation is where the exception arose in my situation, which is why this test ends here. 
I will add some further testing.

1.  I do not think so. At this point, I only ensured that the existing tests still succeeded and that no exception was thrown on compilation (my use case). 
Also, I am not quite sure about what you mean with using these properties as constructor parameters.

DocSvartz:
@JMolenkamp  I meant that if the constructor parameters in your cases can have names with dots, then you should pay special attention to this.


`var destination = new ( source."Some.Property.With.Periods");`

JMolenkamp:
@DocSvartz Got it. It won't be an issue in my case, the dynamically created types have a default constructor and each property both has a getter and a setter. I might take a look though.

JMolenkamp:
@DocSvartz 
Added some tests:
- Mapping to and from a property containing a period does work
- Mapping to a constructor parameter containing a period does work
- Configuring a mapping using a property path string does not work, as it is still split on the periods
- Mapping to a dictionary key containing periods does not work, also didn't work originally. This one is probably not really a part of this PR, but I wanted to take a look at it anyway.

The latter two gave me an idea to try a different approach and keep the period joined string on the InvokerModel. 
Instead of only using the parts after splitting as member names, it might be possible to try out combinations of consecutive parts when members for singular parts are not found. Not really sure what this would involve, but I might take a shot.

src/Mapster.Tests/WhenMappingMemberNameContainingPeriod.cs
```diff
@@ -0,0 +1,298 @@
...
+            Expression<Func<TWithProperty, int>> getPropertyExpression = BuildGetPropertyExpression<TWithProperty, int>(MemberName);
+            Func<TWithProperty, int> getProperty = getPropertyExpression.Compile();
+
+            // Verify
+            Assert.AreEqual(value, getProperty.Invoke(adaptedSource));
+        }
+    }
+
+    [TestMethod]
+    public void Using_Property_Path_String_Is_Supported()
```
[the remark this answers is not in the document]
JMolenkamp: Yes

JMolenkamp:
@DocSvartz 
I tried the alternative approach which turns out to be less impactful and might be preferred.
See [this branch](https://github.com/JMolenkamp/Mapster/compare/development...JMolenkamp:Mapster:feat-map-members-containing-periods)

src/Mapster.Tests/WhenMappingMemberNameContainingPeriod.cs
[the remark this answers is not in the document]
DocSvartz: @JMolenkamp 
I managed to get this test to work.
The only thing that doesn't work is the first mapping.

src/Mapster.Tests/WhenMappingMemberNameContainingPeriod.cs
```diff
@@ -0,0 +1,298 @@
...
+            .NewConfig(typeof(Source), targetType)
+            .Map(MemberName, nameof(Source.Value));
+        TypeAdapterConfig
+            .GlobalSettings
+            .NewConfig(targetType, typeof(Source))
+            .Map(nameof(Source.Value), MemberName);
+
+        // Execute the mapping both ways
+        Source source = new() { Value = 551 };
+        object target = source.Adapt(typeof(Source), targetType);
```
DocSvartz, line 122: this not working. 
551 not Not transmitted

DocSvartz:
```
 public static Expression PropertyOrFieldPath(Expression expr, string[] path)
 {
     if(expr.Type.IsAutoClass && path.Length > 1)
     {
        return Expression.PropertyOrField(expr, string.Join(".", path));
     }
     return path.Aggregate(expr, PropertyOrField);
 }
```

src/Mapster.Tests/WhenMappingMemberNameContainingPeriod.cs
JMolenkamp: I know, that was what I was trying to tell [here](https://github.com/MapsterMapper/Mapster/pull/777/files/0b9b05fc502d656f1107edb16bd69988dc4b007b#r1974020410)

DocSvartz: Without this fix

```
public static Expression PropertyOrFieldPath(Expression expr, string[] path)
 {
     if(expr.Type.IsAutoClass && path.Length > 1)
     {
        return Expression.PropertyOrField(expr, string.Join(".", path));
     }
     return path.Aggregate(expr, PropertyOrField);
 }
```
i get this error

source=Types.Target
destination=Mapster.Tests.WhenMappingMemberNameContainingPeriod+Source
type=Map ---> System.ArgumentException: 'Some' is not a member of type 'Types.Target' (Parameter 'propertyOrFieldName').

JMolenkamp: I know (about the error at least), it gave me an idea for another approach, see [this branch](https://github.com/JMolenkamp/Mapster/compare/development...JMolenkamp:Mapster:feat-map-members-containing-periods) or see my post above where I mentioned it.
All my issues are fixed by only changing the method you mention here.

DocSvartz: As far as I understand. You just need to separate
 "Toplevel" (root) Member Access from "Insider" (nested) Member Access.

```
this "Some.Property.With.Periods" - is top level Member
 target."Some.Property.With.Periods" 

not nested
          target.Some.Property.With.Periods

target
{
 Some {get;set} 
}
where
      Some
         {
               Property {get; set;}
                    
         }
where 
     Property
     {
         With {get; set;}
     }
    . .... and so on
```

here actually requires analysis of mapping line
But for Toplevel, This kind of processing will be quite sufficient.

```
public static TSetter Map<TSetter>(
    this TSetter setter, string destinationMemberName, string sourceMemberName) where TSetter : TypeAdapterSetter
{
    setter.CheckCompiled();

   var x =  setter.Config.RuleMap.First().Key;

    var isTopLevelMemberSourceName = x.Source.GetFieldsAndProperties().Any(x => x.Name == sourceMemberName);
    var isTopLevelMemberDestName = x.Destination.GetFieldsAndProperties().Any(x => x.Name == destinationMemberName);

    setter.Settings.Resolvers.Add(new InvokerModel
    {
        DestinationMemberPath = isTopLevelMemberDestName ? new[] { destinationMemberName } : destinationMemberName.Split('.'),
        SourceMemberPath = isTopLevelMemberSourceName ? new[] { sourceMemberName } : sourceMemberName.Split('.'),
        Condition = null
    });
    return setter;
}
```
I'll think about it some more.

DocSvartz: @JMolenkamp  I saw your solution.
I'm just not really a fan of basing processing on exceptions (it is slow). 

If you don't plan to add any fixes for non-working tests to this PR, add an [Ignore] annotation for them.

DocSvartz: @JMolenkamp If #781 solves your problem and all tests pass.
Then move this PR to draft.

Unfortunately, I don't have time to evaluate the possible application of your PR to solve other problems.
Therefore, I have no preferences regarding one solution or another.

But I don't see any problems with adding it if it doesn't change the current behavior for other cases.

JMolenkamp:
Moved to draft in favor of #781

Issue #783, opened 2025-03-20
RequireExplicitMapping is not working as expected.


JasonGrass:
When converting between two class, both of which have a property named `MyProperty` with different types (one is string and the other is int), 
I have configured `RequireExplicitMapping = true`.

I expected an exception to occur during the `Compile` phase, indicating that I should explicitly configure the conversion between these two properties. However, that did not happen. The error only occurs at runtime if the conversion fails.

Is there any way to avoid this situation?

Mapster version: 7.4.0

```cs
// PatientInformation
public string MyProperty { get; set; } = "";

// PatientInfoViewObject
public int MyProperty { get; set; }
```

```cs
  public static void ConfigureMappings()
  {
      TypeAdapterConfig.GlobalSettings.RequireExplicitMapping = true;
      TypeAdapterConfig.GlobalSettings.RequireDestinationMemberSource = true;

      ConfigurePatientInfoMappings();

      //  expect an exception here
      TypeAdapterConfig.GlobalSettings.Compile();

      // test code
      var info = new PatientInformation();
      info.MyProperty = "test";
      var viewObject = info.Adapt<PatientInfoViewObject>();
  }

  private static void ConfigurePatientInfoMappings()
  {
      TypeAdapterConfig<PatientInformation, PatientInfoViewObject>.NewConfig();

      TypeAdapterConfig<PatientInfoViewObject, PatientInformation>.NewConfig();
  }
```

DocSvartz:
@JasonGrass now   `RequireExplicitMapping = true` work only  for mapping  Class to Class.

Tou create custom config here ` private static void ConfigurePatientInfoMappings()` That's why it doesn't work.
It doesn't work for primitive types.

This needs some thought, but could be expanded.

DocSvartz:
@JasonGrass  If your goal is to make sure that all properties are mapping, then you can use the following config

```
TypeAdapterConfig<Source783, Destination783>.NewConfig()
    .IgnoreNonMapped(true);
```

But then you have to manually map all the Properties. 

But as I understand it, this is not quite what you need?

DocSvartz:
@andrerav  @stagep can add the following setting:

`  TypeAdapterConfig.GlobalSettings.RequireExplicitMappingPrimitive = true; `

[...]

PR #784, opened 2025-03-22
Fix Issue #783 - Add new Feature RequireExplicitMappingPrimitive to config

DocSvartz:


DocSvartz:
I'll add a hint from the configuration and Wiki later.
I still need to describe it :)

Issue #414, opened 2022-03-22
NullReferenceException when mapping from multiple sources when one source is null


jkuek:
I have 3 source items to map to a destination. There is a possibility the third item will be null.

I create a config as follows:

```
            var config = TypeAdapterConfig<(a, b, c), Output>.NewConfig()
                .Map(dest => dest, src => src.Item1)
                .Map(dest => dest, src => src.Item2)
                .Map(dest => dest.Application, src => src.Item3 == null ? (Application) null : new Application()
                {
                    Id = src.Item3.Id,
                    Name = src.Item3.Name
                })

```

My expectation was that a null value for Item3 should be handled and a null value returned for the Application property of the output.

When I run the mapping:
`
var result = (x, y, z)Adapt<Output>(config.Config);
`
when z is null, then a NullReferenceException is thrown.

Is this a bug?

andrerav:
At first glance, this looks like a bug. I will attempt to reproduce this with a regression test and see if I can pinpoint the problem. Is it possible for you to post a stack trace, or relevant parts of a stack trace? Can you also confirm that you are using the latest version of Mapster?

jkuek:
I confirm this issue is occurring using the Mapster 7.3.0 Nuget package.

The behaviour when using a tuple with a single element matches my expectations. For instance:

```
source.Adapt<Dest>() returns null if source is null

(source).Adapt<Dest>() returns null if source is null
```

but when using multiple elements it throws an exception (even without a config specified)

`(source1, source2).Adapt<Dest>() throws NullReferenceException
`

0xSa9:
Yes! I have same problem with version 7.3.0

` config.ForType<(Pair?, Record?, Favorite?), PairDto>()
              .Map(dest => dest, src => src.Item1)
              .Map(dest => dest, src => src.Item2, srcCond => srcCond.Item2 != null)
              .Map(dest => dest, src => src.Item3, srcCond => srcCond.Item3 != null);`

DocSvartz:
A class as a reference type can always have a Null value.
NullableExpressionVisitor only checked for the presence of NullableAttribute

DocSvartz:
This should be fixed in new release versions.
If your issue is still active, please reopen it or create new issue. Thanks.

PR #788, opened 2025-04-09
Fix Issue #414 -  NullableExpressionVisitor improvement

DocSvartz:
and test

### DIFF
--- a/src/Mapster.Tests/WhenMappingNullablePrimitives.cs
+++ b/src/Mapster.Tests/WhenMappingNullablePrimitives.cs
@@ -1,6 +1,6 @@
-﻿using System;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Shouldly;
+using System;
 
 namespace Mapster.Tests
 {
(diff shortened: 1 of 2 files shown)


stagep:
@DocSvartz @andrerav 
Does the exception occur because the mapping attempts to map the properties of PeopleDto which is null? My expectation is that mapping of a reference type that is null stops at the null type and does not access the properties of the null type avoiding a NullReferenceException. Is your question asking if this behavior should be opt-in through an attribute on the potentially nullable type or should it be automatic?

Looking at my own code generated mappings, it always checks the reference type is not null before it proceeds to map properties of the reference type. Is this the behavior that you are referring to that was implemented "by mistake"?

DocSvartz:
> Does the exception occur because the mapping attempts to map the properties of PeopleDto which is null? 
What matters here is whether the Source Properties can be null.

code which is currently being created from first case:
```
    .If ($var1 == null) {
        null
    } .Else {
        .New MapsterTest3.Test1+UserDto(){
            Id = $var1.Id,
            People = .If ($var1.People == null) { // Property People not Mark Nullable attribute and now Mapster considers this property maybe null
                null
            } .Else {
                .New MapsterTest3.Test1+PeopleDto(){
                    Id = ($var1.People).Id,
                    Name = ($var1.People).Name
                }

```
code which is currently being created from second case:
```
  .If ($var1 == null) {
        null
    } .Else {
        .New MapsterTest3.Test1+UserDto(){
            Id = $var1.Id,
            Name = $var1.Name,
            People = .New MapsterTest3.Test1+PeopleDto(){  // Property People automatic Mark Nullable[1]
                Id = ($var1.People).Id,
                Name = ($var1.People).Name
            }
```

> Is your question asking if this behavior should be opt-in through an attribute on the potentially nullable type or should it be automatic?

The problem here is that both the first and second options when describing a type in a nullable aware context (#nullable enable) have never been null. 

> [A variable of a reference type T must be initialized with non-null, and can never be assigned a value that might be null.](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types)

And NullabilityInfoContext will indeed define it as not null in both cases. It can also take into account other nullable attributes.

> Looking at my own code generated mappings, it always checks the reference type is not null before it proceeds to map properties of the reference type. Is this the behavior that you are referring to that was implemented "by mistake"?

If you mean Mapster Tool. Then source generators work outside of nullable aware context
and there this behavior is correct.

Viewing the code of what happens during normal mapping is better done like this user.BuildAdapter().CreateMapExpression<UserDto>();

DocSvartz:
@stagep @andrerav  Mapster for the most part, it does not analyze the runtime state of variables. It creates mapping functions based on the Source and Destination types.
From this point of view, if we are provided with types whose members are explicitly specified as non-nullable, then it is quite expected that the processing function will be created for types that do not allow null values.
(In this case, the presence of null is already an error on the calling side - a violation of the established contract.)

I am not against the existence of behavior "by design", That's why I'm asking what we will stick to.

stagep:
@DocSvartz Have been busy so not got to look at this. What version or branch of Mapster are you using in the above example? I see you have forked this repository so do I need to use a branch of your repository?

DocSvartz:
@stagep  everything I described happens on any version.
If you need a fixed version, you can use this PR. It was created on the current Development branch
Or do you need a fix based on NullabilityInfoContext ?

**Current tests do not use nullable context. To notice this, you need to use `#nullable enable` in the file with the description of test classes.**

Issue #789, opened 2025-04-11
Generic class mapping with inheritance


luczito:
I am in the process of refactoring a project from automapper to mapster. 
I have run into an issue with mapping configs for generic classes e.g.

```C#
CreateMap(typeof(derivedDto<>), typeof(derivedDomain<>))
        .IncludeBase(typeof(baseDto), typeof(baseDomain)
        .ReverseMap();
```

As far as i can tell i can create a config for the base class via mapster as:

```C#
TypeAdapterConfig.GlobalSettings.ForType(typeof(derivedDto<>), typeof(derivedDomain<>))
    .Map("value", "Value");
```

But i (as far as i can tell) can not use .Include or .Inherits to include the inheritance information in the mappings.

Is this possible somehow or do i need to create a custom mapper to achieve this?

DocSvartz:
@luczito  This is not included in the standard distribution because there is no way to determine that you are actually copying settings for types within same inheritance hierarchy.

You can Create Custom Helper. 

**Limitations and warnings when using:**

1. You should check for yourself that they are actually used for types in the same inheritance hierarchy. 
2. This may stop working (in future versions) or work incorrectly.

```

public static TSetter Inherits<TSetter>(this TSetter setter, Type TBaseSource, Type TBaseDestination) where TSetter : TypeAdapterSetter
 {
     setter.CheckCompiled();

     if (setter.Config.RuleMap.TryGetValue(new TypeTuple(TBaseSource, TBaseDestination), out var rule))
     {
         setter.Settings.Apply(rule.Settings);
     }
   
     return setter;
 }

public static TSetter Include<TSetter>(this TSetter setter, Type sourceType, Type destType) where TSetter : TypeAdapterSetter
{
    setter.CheckCompiled();

    setter.Config.Rules.LockAdd(new TypeAdapterRule
    {
        Priority = arg =>
            arg.SourceType == sourceType &&
            arg.DestinationType == destType ? (int?)100 : null,
        Settings = setter.Settings
    });

    setter.Settings.Includes.Add(new TypeTuple(sourceType, destType));

    return setter;
}



```

DocSvartz:
This should be fixed in new release versions.
If your issue is still active, please reopen it or create new issue. Thanks.

PR #790, opened 2025-04-12
Fix Issue #789 Add support .Inherits() and .Include() to map OpenGenerics

DocSvartz:


DocSvartz:
Defining the type as OpenGenerics is most likely not very reliable.
It looks like it might work) 

Additional samples are needed for testing.

Issue #820, opened 2025-10-13
Mapster.EFCore ProjectToType<T>().FirstOrDefaultAsync throwing exception


ScarletKuro:
Very similar problem as here https://github.com/MapsterMapper/Mapster/issues/566
But I don't think that "materializing the query" is an acceptable workaround:

This **WORKS** (without using `IMapper`)
```C#
var template = await query
	.ProjectToType<TemplateDto>()
	.FirstOrDefaultAsync();
```
This **WORKS** (with `IMapper` but `FirstOrDefault`)
```C#
var template = _mapper.From(query)
	.ProjectToType<TemplateDto>()
	.FirstOrDefault();
```
This **doesn't WORK** (With `IMapper` but `FirstOrDefaultAsync`)
```C#
var template = await _mapper.From(query)
	.ProjectToType<TemplateDto>()
	.FirstOrDefaultAsync();
```

And thows:
```
System.MissingMethodException
  HResult=0x80131513
  Message=Constructor on type 'Mapster.EFCore.MapsterAsyncEnumerable`1[[MessageDispatcher.Shared.Restful.Api.Dto.TemplateDto, MessageDispatcher.Shared, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]' not found.
  Source=System.Private.CoreLib
  StackTrace:
   at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture)
   at System.Activator.CreateInstance(Type type, Object[] args)
   at Mapster.EFCore.MapsterQueryableProvider.ExecuteAsync[TResult](Expression expression, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions.ExecuteAsync[TSource,TResult](MethodInfo operatorMethodInfo, IQueryable`1 source, Expression expression, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions.ExecuteAsync[TSource,TResult](MethodInfo operatorMethodInfo, IQueryable`1 source, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions.FirstOrDefaultAsync[TSource](IQueryable`1 source, CancellationToken cancellationToken)
[...]

Issue #566, opened 2023-03-17
Mapster.EFCore ProjectToType<T>().SingleOrDefault throwing exception


MortenRickiRasmussen:
Hi!

I have this piece of code 

``` c#
public async Task<T> SingleOrDefault(int id) 
{
  var query = DbContext.SomeEntity.Where(x => x.Id == id);

   return await _mapper.From(query).ProjectToType<T>().SingleOrDefaultAsync();
}
```

it throws the following Exception 

```
Data: System.MissingMethodException: Constructor on type 'Mapster.EFCore.MapsterAsyncEnumerable`1[[SomeSystem.SomeDto, SomeSystem.Api, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]' not found.
   at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture)
   at Mapster.EFCore.MapsterQueryableProvider.ExecuteAsync[TResult](Expression expression, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions.ExecuteAsync[TSource,TResult](MethodInfo operatorMethodInfo, IQueryable`1 source, Expression expression, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions.ExecuteAsync[TSource,TResult](MethodInfo operatorMethodInfo, IQueryable`1 source, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions.SingleOrDefaultAsync[TSource](IQueryable`1 source, CancellationToken cancellationToken) .........
```

If i exchange the code to the following, it works fine

``` c#
public async Task<T> SingleOrDefault(int id) 
{
   var query = DbContext.SomeEntity.Where(x => x.Id == id);

   var list = await _mapper.From(query).ProjectToType<T>().ToListAsync();

   return list.FirstOrDefault();
}
```

Additional info:

AspNet 7.0
Mapster 7.3.0
Mapster.EFCore 5.1.0

andrerav:
Have you tried materializing the list before calling ProjectToType()? There are some issues when using ProjectToType with async queries.

andrerav:
Assuming this was solved by materializing the query first.

SeiOkami:
Looks like a bug

[...]

PR #821, opened 2025-10-14
Mapster.EF.Core: Add type check for IAsyncEnumerable<>

ScarletKuro:
Fixes: https://github.com/MapsterMapper/Mapster/issues/820
Fixes: https://github.com/MapsterMapper/Mapster/issues/566 (was closed due to a workaround)

Description:
The issue occurs because the code always attempts to wrap the result in `MapsterAsyncEnumerable`, even when the type does not implement `IAsyncEnumerable`. This fix ensures wrapping only happens if the type actually implements `IAsyncEnumerable`, which prevents the `MissingMethodException` to throw.

### DIFF
--- a/src/Mapster.EFCore.Tests/EFCoreTest.cs
+++ b/src/Mapster.EFCore.Tests/EFCoreTest.cs
@@ -1,11 +1,12 @@
-using System;
-using System.Collections.Generic;
-using System.Linq;
 using Mapster.EFCore.Tests.Models;
 using MapsterMapper;
 using Microsoft.EntityFrameworkCore;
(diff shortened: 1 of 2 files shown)


DocSvartz:
@ScarletKuro Can you do this using the development branch?
So we can include it in the next pre-releases version.

ScarletKuro:
> @ScarletKuro Can you do this using the development branch?
> So we can include it in the next pre-releases version.

Sure, do I leave this open tho?

DocSvartz:
Yes, leave this open.

DocSvartz:
Mergeted to Develop branch

Issue #138, opened 2018-02-07
The type initializer for 'Mapster.TypeAdapter`2' threw an exception.


mvareeth:
Hi

I am using Mapster 3.0.4 with DotNet Core 1.0. I am getting following intermittent error in production where the application has many user bases. I doubt Mapster is not considering thread safety while converting the object model.  Can you please help to solve this issue?

"System.TypeInitializationException: The type initializer for 'Mapster.TypeAdapter`2' threw an exception. ---> System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Enumerator.MoveNext()
   at Mapster.SettingStore.Apply(SettingStore other)
   at Mapster.TypeAdapterConfig.GetMergedSettings(TypeTuple tuple, MapType mapType)
   at Mapster.TypeAdapterConfig.GetCompileArgument(TypeTuple tuple, MapType mapType, CompileContext context)
   at Mapster.TypeAdapterConfig.CreateMapExpression(TypeTuple tuple, MapType mapType)
   at Mapster.TypeAdapterConfig.<GetMapFunction>b__55_0(TypeTuple tuple)
   at Mapster.TypeAdapterConfig.AddToHash(Hashtable hash, TypeTuple key, Func`2 func)
   at Mapster.TypeAdapterConfig.GetMapFunction(Type sourceType, Type destinationType)
   at Mapster.TypeAdapterConfig.GetMapFunction[TSource,TDestination]()
   at Mapster.TypeAdapter`2..cctor()
   --- End of inner exception stack trace ---
   at Mapster.TypeAdapter.Adapt[TSource,TDestination](TSource source)

Thanks

chaowlert:
TypeAdapterConfig is not thread self. Please setup config on application start only.

If you really would like to keep config with mapping, pls use `Fork` to create new config (https://github.com/MapsterMapper/Mapster/wiki/Config-location#keep-together-with-mapping).

Issue #86, opened 2016-10-14
thread-safety issue with TypeAdapterConfig


centur:
Today after running a large suit of tests at once I saw one test failed with this exeption

```
System.ArgumentException
Destination array was not long enough. Check destIndex and length, and the array's lower bounds.
   at System.Array.Copy(Array sourceArray, Int32 sourceIndex, Array destinationArray, Int32 destinationIndex, Int32 length, Boolean reliable)
   at System.Collections.Generic.List`1.CopyTo(T[] array, Int32 arrayIndex)
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.<ReverseIterator>d__74`1.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.OrderedEnumerable`1.<GetEnumerator>d__1.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Mapster.TypeAdapterConfig.GetMergedSettings(Type sourceType, Type destinationType, MapType mapType)
   at Mapster.TypeAdapterConfig.CreateMapExpression(Type sourceType, Type destinationType, MapType mapType, CompileContext context)
   at Mapster.TypeAdapterConfig.CreateInlineMapExpression(Type sourceType, Type destinationType, MapType mapType, CompileContext context)
   at Mapster.Adapters.BaseAdapter.CreateAdaptExpression(Expression source, Type destinationType, CompileArgument arg)
   at Mapster.Adapters.CollectionAdapter.CreateArraySet(Expression source, Expression destination, CompileArgument arg)
   at Mapster.Adapters.BaseAdapter.CreateBlockExpressionBody(Expression source, Expression destination, CompileArgument arg)
   at Mapster.Adapters.BaseAdapter.CreateExpressionBody(Expression source, Expression destination, CompileArgument arg)
   at Mapster.Adapters.BaseAdapter.CreateAdaptFunc(CompileArgument arg)
[...]

PR #827, opened 2025-11-06
Fix Race Condition between Configuration and Adapt()

DocSvartz:
Based on[ these tests](https://github.com/MapsterMapper/Mapster/blob/master/src/Mapster.Tests/WhenRegisteringAndMappingRace.cs) and #86, #138

Depending on the speed and order of execution (configuration and Adapt()), one of the following situations can occur:
1) Incomplete configuration (the main cause of errors in the tests mentioned at the beginning);
2) "Incomplete cleanup" - Exception: "TypeAdapter.Adapt was already called, please clone or create a new TypeAdapterConfig."
3) Exception to collection modification during enumeration.

Additional problem:
The current configuration design does not provide information about configuration completion.

add This PR 

1. Thread safety creating a new configuration:
```
 TypeAdapterConfigConcurrency<WhenAddingCustomMappings.SimplePoco, WeirdPoco>
                .NewConfig(cfg =>
                {
                    cfg
                        .Map(dest => dest.IHaveADifferentId, src => src.Id)
                        .Map(dest => dest.IHaveADifferentId, src => src.Id)
                        .Map(dest => dest.MyNamePropertyIsDifferent, src => src.Name)
                        .Ignore(dest => dest.Children);
                });

```

2.  Thread safety loading configurations:

`TypeAdapterConfig.GlobalSettings.ScanConcurrency(Assembly.GetExecutingAssembly());`

### DIFF
--- a/src/Mapster.Tests/WhenRegisteringAndMappingRace.cs
+++ b/src/Mapster.Tests/WhenRegisteringAndMappingRace.cs
@@ -39,6 +39,12 @@ namespace Mapster.Tests
 
             var simplePoco = new WhenAddingCustomMappings.SimplePoco {Id = Guid.NewGuid(), Name = "TestName"};
 
+            //first state (i = 0) Must be configured
+            TypeAdapterConfig<WhenAddingCustomMappings.SimplePoco, WeirdPoco>.NewConfig()
+                               .Map(dest => dest.IHaveADifferentId, src => src.Id)
+                               .Map(dest => dest.MyNamePropertyIsDifferent, src => src.Name)
+                               .Ignore(dest => dest.Children);
+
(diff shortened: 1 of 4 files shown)


DocSvartz:
@andrerav  @stagep See if this is necessary at all? 
Actually, I don't fully understand the meaning of the initial tests :)

DocSvartz:
This problem can perhaps be partially solved using Freezing config (without throwing exceptions when changing). However, this needs to be tested with at least 1 million iterations.

But maybe I'm not quite understanding the real world use case for this.

The only real use case I see is when different adapt configurations are set in different threads.
But in that case, the Adapt call itself must be called on the same thread. (This should be a single lock, covering both the configuration and the Adapt call.)

PR #845, opened 2026-01-05
Merge Development to Master - For work on documentation

DocSvartz:
Sync development and master so you can work on new documentation


DocSvartz:
@andrerav  @stagep  @DevTKSS  I suggest synchronizing the branches at this stage so we can start working on the documentation independently of the code. What do you think about this?

stagep:
> @andrerav @stagep @DevTKSS I suggest synchronizing the branches at this stage so we can start working on the documentation independently of the code. What do you think about this?

Agree

andrerav:
> @andrerav @stagep @DevTKSS I suggest synchronizing the branches at this stage so we can start working on the documentation independently of the code. What do you think about this?

Sounds good to me :)

DocSvartz:
On the second try the site seemed to launch.  😅

image

@DevTKSS  I'm not sure this is the correct setting. But it was originally configured for the gh-pages branch, and [publishing failed](https://github.com/MapsterMapper/Mapster/actions/runs/20718384748/attempts/1). 🤔

DevTKSS:
@DocSvartz I could send you a screenshot of my settings in my Samples Repo. Do you still need this or did it work now? I am not at my computer right now

The CI file and `docfx.json` is already public available in there

https://github.com/DevTKSS/DevTKSS.Uno.SampleApps/blob/master/.github%2Fworkflows%2Fbuild-deploy-docs.yml

Did just recently push some new docs releases, which worked fluentles 👍 just mine includes metadata with filterconfig.yml because I don't want the generated Mvux API of my app to be provided there, just because it's ridiculous to assume or expect externals to make calls to generated VM code 🤣 or Uno own API like the Resizetizer 😅 that's where you will find differences to Mapster's version now

DocSvartz:
I don't know, but it seems to have already worked.

 https://mapstermapper.github.io/Mapster/ is active
And it's published in the same way as your Samples Repo docs  to special github Environment.

DocSvartz:
Actually, I wanted to know if you have the same publishing setting set up in your Samples Repo? 
If not, then two things remain unclear:
1) Was it the workflow you added that triggered it, or something else?
2) Will the site automatically rebuild after changes to the master branch? 
😅🤔

DevTKSS:
@DocSvartz GitHub does always take some time between pages deployment and the actual update of the page we can explore. Most likely you have been to quick for GitHub possibly?👍

DocSvartz:
No, initially, the site build didn't work. 
After changing the settings to those shown in the image and manually restarting the build process, the site was built and working.

DevTKSS:
@DocSvartz yes, I would say, looks same like I have:
image

I am still curious how to improve my knowledge about the environments thing, but the github docs did not really made it better. Maybe some day I will find someone who knows about this and can ask this one 👍

DocSvartz:
Ok.  That's what I wanted to know. 👍

PR #856, opened 2026-03-06
Fix docs - Records type description to Mapster v10

DocSvartz:


DocSvartz:
@andrerav  @stagep   In my opinion, this is a description of the most critical changes (which do not need to be activate manual).

If the description needs to be corrected, please let me know or confirm that it can be merge.

stagep:
I would change these 2 lines for clarity

100 > Mapster treats the Record type as an immutable type.

Added the

101 > Only a non-destructive mutation is returned using the with-expression.

More direct language, and using the correct terminology for using with { ... } syntax. I am also using returned instead of available as the mutated record is returned.

Agree?

DocSvartz:
What I wanted to convey is that the result will be similar to how _with expression_ works.
Because you can't call a real _with statement_ through Expression

PR #867, opened 2026-03-18
Replace TFM for Mapster.EF6

DocSvartz:


andrerav:
@DocSvartz Just to check, this wouldn't compile with netstandard2.0 right?

DocSvartz:
@andrerav yes, in ~~netstandard2.0 not like there are no Entity Framework entities there.~~

DocSvartz:
@andrerav  EF6 is not available for netstandard2.0, only netframework and netstandard2.1.

andrerav:
That must be some kind of oversight from Microsoft I think. If they could target netstandard2.1, then surely they could target netstandard2.0 too. But it is what it is :)

PR #870, opened 2026-03-18
fix: Restore PackageLicense

### LINKED ISSUE #869: license missing for package Mapster 10.0.0
Hi,

after updating the main package to the latest version (Mapster 10.0.0) our security scans fail due to a missing license:

> Source license not available; no declaration available is in violation for the following reason(s):
Found 'Not Supported' license
Found 'Not Declared' license

On NuGet there is indeed no license present currently. All other Mapster packages seem to be fine. Would be great if you can fix it.

Kind regards,
Sascha

DocSvartz:
- Drop Mapster/Directory.Build.props


DocSvartz:
@andrerav  @stagep  It is completely unclear how these files work.

In one case, they overwrite what's not there. In the other, it's the opposite.😅

In both cases, nothing works. 😒

This restores the package's license. And it doesn't seem to break the benchmark.development.

PR #884, opened 2026-03-22
Fix #883 - Add support for activating a class constructor using default parameter values + new implementation config params

### LINKED ISSUE #883: v10 no longer uses default values for missing constructor parameters
Considering these 2 classes:

```
public class Source
{
  public required string Value1 { get; init; }
}

public class Target
{
  public Target(string value1, string value2)
  {
    Value1 = value1;
    Value2 = value2;
  }

  public string Value1 { get; }

  public string Value2 { get; }
}
```

Running this code:
```
    var source = new Source
    {
      Value1 = "123",
    };

    var target = source.Adapt<Target>();
```

This works in v7.4.0 resulting in a Source instance with Value1 = "123" and Value2 = null.

In v10 I get this exception:
System.InvalidOperationException: No default constructor for type 'Target', please use 'ConstructUsing' or 'MapWith'

Sounds like a pretty huge breaking change in my opinion. I dont see anything about it in the changelog.

DocSvartz:


DocSvartz:
- not correct implementation record detector

PR #889, opened 2026-03-23
Improvement build - part 1

### LINKED ISSUE #843: Unification build Properties
@DocSvartz @andrerav Seeing all the repeating of Properties here I would like to suggest to:
- use `Directory.Packages.props` -> introduce CPM
  
  - only one place we need to update
  - Configurable along with the Variables of `Directory.Build.props` if you want to ensure to use specific versions of packages for projects like Source Generators or Test Projects (which by the way are missing the common `IsTestProject` Tag, which I added as suggestion to this PR directly)

- use `Directory.Build.props`
  - Define NuGet Package Information Centrally - e.g. I found this here and copied it to [my own Repo + adjusted the Properties (Name, License etc.)](https://github.com/DevTKSS/DevTKSS.MyManufacturerERP/blob/d7fbc0f5aef82f526dd7e1e17a5f73dd4ee33d42/Directory.Build.props) I dont use it by now, so adding some other references:
    - [First Example](https://github.com/unoplatform/uno.extensions/blob/main/src/Directory.Build.props)
    - [Second Example](https://github.com/unoplatform/uno/blob/master/Directory.Build.props)
  
  - source regular Project Props we need repentantly using Variables for example:
    - `Directory.Build.props`
    
      ```xml
      <Project>
        <PropertyGroup>
          <ImplicitUsings>enable</ImplicitUsings>
          <Nullable>enable</Nullable>
          <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
          <DefaultTargetFrameworks>net10.0;net9.0;net8.0</DefaultTargetFrameworks>
          <Authors>Some Name;Some OtherName</Authors>
        </PropertyGroup>
      </Project>
      ```
     - Regular `*.csproj`
     
       ```xml
       <Project Sdk="Microsoft.NET.Sdk">
          <PropertyGroup>
            <TargetFrameworks>$(DefaultTargetFrameworks)</TargetFrameworks>
            <RootNamespace>Mapster</RootNamespace>
            <AssemblyName>Mapster</AssemblyName>
            <Authors>$(Authors)</Authors>
            <OutputType>Library</OutputType>
          </PropertyGroup>
        </Project>
        ```

- Update the Authors Entries?

  Never seen those Names recently and @andrerav mentioned in the discussion, he is the only Maintainer except from @DocSvartz and @stagep (potentially missed one 😄 but you get what I mean hopefully) which I did not see in one of those Lists.

_Originally posted by @DevTKSS in https://github.com/MapsterMapper/Mapster/issues/841#issuecomment-3694104162_

DocSvartz:


DocSvartz:
@andrerav  Is this a bug, or was it really intended that the values ​​from build props were not supposed to be used purposes in the Mapster library?

PR #924, opened 2026-04-21
Demonstrate a record mapping issue

mindysport:
Please note: 
- This PR isn't intended to be merged as is but exists essentially to repro an issue that I found when mapping records under specific configurations. It seemed more helpful / straightforward to demonstrate this via tests as opposed to just explaining it. 
- Essentially, the issue I encountered was backing fields sneaking in when using IgnoreNonMapped and IgnoreNulls in connection with mapping to a pre-existing record, such that the backing fields would be re-applied to the record at the end, overwriting whatever mapping should have taken place.

Changes:
- Added tests to demonstrate an issue mapping records where specific configurations allow backing fields to sneak in.
- Added a (suggested) fix that allows the tests to pass. Both changes (in isolation from each other and on their own) solve the immediate problem and cause the tests to pass, but I'm not sure which fix (if any) fits best with the codebase, but I wanted to attempt to provide a suggestion at least.

### DIFF
--- a/src/Mapster.Tests/WhenMappingRecordRegression.cs
+++ b/src/Mapster.Tests/WhenMappingRecordRegression.cs
@@ -4,994 +4,1456 @@ using System;
 using System.Collections.Generic;
 using System.Text.Json;
 using static Mapster.Tests.WhenExplicitMappingRequired;
+using System.Linq.Expressions;
+using System.Reflection;
 using static Mapster.Tests.WhenMappingDerived;
 
 namespace Mapster.Tests
 {
-    /// <summary>
-    /// Tests for https://github.com/MapsterMapper/Mapster/issues/537
-    /// </summary>
-    [TestClass]
-    public class WhenMappingRecordRegression
-    {
-        [TestMethod]
-        public void AdaptRecordToRecord()
-        {
-            TypeAdapterConfig<TestRecordY, TestRecordY>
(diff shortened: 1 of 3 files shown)


DocSvartz:
Hi @mindysport.
 Based on your description, I have a rough idea of ​​what this might be related to. ~~But the examples you provided shouldn't and don't generate access to backing fields.~~

Perhaps you have a UseDestinationValue setting in your global configuration with some criteria. Or do the fields themselves have this attribute?

Update: This does not appear in all examples.

mindysport:
@DocSvartz You're right that it doesn't appear in all the example tests I created, and I should have been clearer with them (perhaps I should have removed all of them except the ones that fail, but I thought the extra tests helped narrow the scope of the issue I'm seeing). The tests demonstrate that the failure case is ONLY in connection with IgnoreNonMapped(true), so for me, only the 4 tests where the config looks something like this are failing: 
image

Also, because I'm passing the created config in the scope of the tests, I don't think it's related to a global config setting (I created these repro examples in tests to try to rule out my own custom configurations).

As a specific example, here's what the expression string generated by CreateMapToTargetExpression looks like for me at the end of the MapToTargetWithIgnoreNonMappedShouldNotContainBackingFields test (when I turn off the suggested fixes and revert the source code to its original state):
image

PR #949, opened 2026-05-18
[Go to v10.0.8] - fix mapster version in benchmark result

DocSvartz:

PR #951, opened 2026-05-18
[Go to v10.0.8] Push to master branch - Release v10.0.8

### LINKED ISSUE #984: Adapt to existing object not work when have readonly property
Please help.
```c#
void Main()
{
	var a1=new Foo(){Id=3,Name="aa"};
	var a2=new Foo(){Id=3};
	a1.Adapt(a2);
	a2.Name.Dump(); // will got null
}

class Foo(){
	public int Id {get;set;}
	public string Name{get;set;}
	public string Prop {get;} //when have setter, no problem
}
```

DocSvartz:
Can be sent after merging:
 - #948 
 - #949 
 - #950 
 - #956 
 - #957
 - #981

 - #972

Maybe: 
 - #969
- #979


DocSvartz:
Hi, @andrerav  @stagep.  Please review and approve if everything is good.

DocSvartz:
The Commisar error is related to commit names. I couldn't find a way to change their names using `git cp` and still preserve the history.

DocSvartz:
Okay, since there are no obvious objections and no new bug reports, I'll merge this myself 😅

PR #960, opened 2026-05-25
fix: skip destination transforms for explicit member maps (#952)

### LINKED ISSUE #952: DestinationTransform.EmptyCollectionIfNull overrides explicit null mapping
When DestinationTransform.EmptyCollectionIfNull is configured globally via config.Default, it still overrides explicitly defined mappings that return null.
This behavior is unexpected because explicit mappings (via .Map(...)) should take precedence over global/default transforms.
 
- Mapster version: 10.0.8-pre06
- .NET version: net10.0

```csharp
class FooDto
{
    public string[]? Strings { get; set; }
}

record Foo(string?[] Strings);

public class DestinationTransformTests
{
    [Fact]
    public void ExplicitNullMapping_ShouldNotBeOverridden_ByDefaultEmptyCollectionTransform()
    {
        // Configure

        config.Default

        config.NewConfig<Foo, FooDto>()
            .Map(d => d.Strings, _ => (string[]?)null);

        // Arrange
        var fooDto = new Foo([]);

        // Act
        var foo = fooDto.Adapt<FooDto>(config);

        // Assert
        foo.Strings.ShouldBeNull();
    }
}
```

## Expected behavior
The explicit mapping:
```csharp
.Map(d => d.Strings, _ => (string[]?)null)
```
should be respected, resulting in foo.Strings == null.

## Actual behavior
foo.Strings is transformed into an empty array, meaning that EmptyCollectionIfNull is applied after the explicit mapping and overrides it.

## Notes
The transform is registered on config.Default, not per-type.
This suggests that global destination transforms are applied at a stage where they override even explicit mapping results.
This makes it impossible to explicitly map a collection property to null when EmptyCollectionIfNull is enabled globally.

wyf027:
## Summary
- Skip global `DestinationTransform` rules when a member has an explicit `.Map(...)` resolver
- Add regression test for `EmptyCollectionIfNull` with an explicit null mapping

Fixes #952

## Test plan
- [ ] CI passes
- [ ] Existing destination transform tests still pass
- [ ] Explicit `.Map(..., _ => null)` keeps null instead of empty collection

### DIFF
--- a/src/Mapster.Tests/WhenPerformingDestinationTransforms.cs
+++ b/src/Mapster.Tests/WhenPerformingDestinationTransforms.cs
@@ -95,6 +95,19 @@ namespace Mapster.Tests
             destination.Set.Count.ShouldBe(0);
         }
 
+        [TestMethod]
+        public void Explicit_Null_Mapping_Is_Not_Overridden_By_EmptyCollectionIfNull()
+        {
+            var config = new TypeAdapterConfig();
+            config.Default.AddDestinationTransform(DestinationTransform.EmptyCollectionIfNull);
+            config.NewConfig<ExplicitNullSource, ExplicitNullDestination>()
+                .Map(d => d.Strings, _ => (string[]?)null);
+
+            var destination = new ExplicitNullSource([]).Adapt<ExplicitNullDestination>(config);
+
+            destination.Strings.ShouldBeNull();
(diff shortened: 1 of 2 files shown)


wyf027:

wyf027:

PR #962, opened 2026-05-25
fix: handle optional value-type constructor defaults (#954)

### LINKED ISSUE #954: CompileException when mapping record with DateTime to class with optional constructor parameter
Mapping from a record containing a DateTime to a class with a constructor that has a default DateTime parameter causes a Mapster.CompileException.

### Environment

Mapster version: 10.0.8-pre06
.NET version: net10.0

### Repro
```csharp
public class DateTimeFooDto
{
    public DateTime Timestamp { get; set; }

    public DateTimeFooDto(DateTime timestamp = default(DateTime))
    {
        this.Timestamp = timestamp;
    }
}

record DateTimeFoo(DateTime Timestamp);

public class DateTimeMappingTests
{
    [Fact]
    public void DateTimeMapping()
    {
        // Arrange
        var src = new DateTimeFoo(DateTime.Today);

        // Act
        var foo = src.Adapt<DateTimeFooDto>();

        // Assert
        foo.Timestamp.ShouldBe(src.Timestamp);
    }
}
```

### Actual behavior
Throws:
```csharp
...
System.ArgumentException: Argument types do not match
   at System.Linq.Expressions.Expression.Constant(Object value, Type type)
...
```

### Expected behavior
Mapping should succeed and foo.Timestamp should equal src.Timestamp.

It looks like there are several workarounds, but my problem is that I'm using a code generator and changing it will affect lot's of dependencies. So I have to use MapWith in such cases.

wyf027:
## Summary
- Fall back to `CreateDefault()` when optional constructor parameters report a null `DefaultValue`
- Fixes compile failures mapping records with `DateTime` members to classes with optional `DateTime` constructor parameters
- Add regression test from issue repro

Fixes #954

## Test plan
- [ ] CI passes
- [ ] `WhenMappingRecordWithOptionalDateTimeConstructor` succeeds

### DIFF
--- a/src/Mapster.Tests/WhenMappingRecordWithOptionalDateTimeConstructor.cs
+++ b/src/Mapster.Tests/WhenMappingRecordWithOptionalDateTimeConstructor.cs
@@ -1,0 +1,31 @@
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+
+namespace Mapster.Tests
+{
+    [TestClass]
+    public class WhenMappingRecordWithOptionalDateTimeConstructor
+    {
+        [TestMethod]
+        public void Record_With_DateTime_Maps_To_Class_With_Optional_DateTime_Constructor()
+        {
+            var source = new DateTimeFoo(DateTime.Today);
+            var destination = source.Adapt<DateTimeFooDto>();
+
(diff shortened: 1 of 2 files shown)


wyf027:

wyf027:

PR #963, opened 2026-05-25
fix: skip abstract base mapping when nested source is null (#928)

### LINKED ISSUE #928: 10.0.7: nullable abstract property maps through abstract base type and throws `Cannot instantiate type`
I upgraded from Mapster `9.0.0-pre01` to `10.0.7`.

After the upgrade, mapping an object that contains a nullable abstract/base-type property fails when the source property value is `null`.

The mapping is configured with `Include<TDerivedSource, TDerivedDestination>()`, so I would expect Mapster to use the included derived mapping when the source value is non-null, and simply map `null` to `null` when the source value is null.

## Repro
```csharp
public abstract class DtoBase
{
}

public class DtoDerived : DtoBase
{
    public string Id { get; set; } = null!;
}

public class DtoFoo
{
    public DtoBase? Field { get; set; }
}

public abstract class DomainBase
{
}

public class DomainDerived : DomainBase
{
    public string Id { get; set; } = null!;
}

public class DomainFoo
{
    public DomainBase? Field { get; set; }
}
```
## Global Configuration:
```c#
public static TypeAdapterConfig ConfigureMapster(MappingAssemblyProvider mappingAssemblyProvider)
{
    var config = TypeAdapterConfig.GlobalSettings;

    config.Default.PreserveReference(true);
    config.Default.EnumMappingStrategy(EnumMappingStrategy.ByName);
    config.Default.EnableNonPublicMembers(true);
    config.Default.MapToConstructor(true);
    config.Default.ShallowCopyForSameType(false);

    config.AllowImplicitSourceInheritance = true;
    config.RequireDestinationMemberSource = true;

    foreach (var assembly in mappingAssemblyProvider.GetMapperAssemblies())
    {
        config.Scan(assembly);
    }

    config.Compile();

    return config;
}

public class MappingConfig : IRegister
{
    {
        config.ForType<DtoBase, DomainBase>()
            .TwoWays()
            .Include<DtoDerived, DomainDerived>();
    }
}
```
## Test:
```c#
[TestFixture]
public class RecordMappingTests
{
    [SetUp]
    public void SetUp()
    {
        MappingConfiguration.ConfigureMapster(new MappingAssemblyProvider());
    }

    [Test]
    public void DtoFooWithFieldNull_Maps_To_DomainFoo()
    {
        var dto = new DtoFoo
        {
            Field = null
        };

        var domain = dto.Adapt<DomainFoo>();

        Assert.That(domain.Field, Is.Null);
    }

    [Test]
    public void DtoFooWithField_Maps_To_DomainFoo()
    {
        var dto = new DtoFoo
        {
            Field = new DtoDerived
            {
                Id = "id"
            }
        };

        var domain = dto.Adapt<DomainFoo>();

        Assert.That(domain.Field, Is.Not.Null);
        Assert.That(domain.Field, Is.TypeOf<DomainDerived>());
    }
}
```
The DtoFooWithFieldNull_Maps_To_DomainFoo test fails with: 
```
System.InvalidOperationException : Cannot instantiate type: DomainBase

at lambda_method30(Closure, DtoBase)
at lambda_method32(Closure, Object)
at Mapster.TypeAdapter.Adapt[TDestination](Object source, TypeAdapterConfig config)
at Mapster.TypeAdapter.Adapt[TDestination](Object source)
at RecordMappingTests.RecordMappingTests.DtoFooWithFieldNull_Maps_To_DomainFoo()`
```
This only happens then the adapt call is made, and the compile() call succeeds. It looks like Mapster is attempting to map the nested Field property through the declared base mapping, even when the source value is null.

wyf027:
## Summary
- When mapping a nullable abstract/base property configured with `Include<,>()`, Mapster attempted to adapt through the abstract destination type even when the source value was `null`, throwing `Cannot instantiate type` at runtime
- Short-circuit null sources in `CreateAdaptExpression` before adapting to abstract or non-public-ctor destination types

Fixes #928

## Test plan
- [ ] CI passes (`WhenMappingNullableAbstractWithInclude` covers null and derived cases with `MapToConstructor(true)`)

### DIFF
--- a/src/Mapster.Tests/WhenMappingNullableAbstractWithInclude.cs
+++ b/src/Mapster.Tests/WhenMappingNullableAbstractWithInclude.cs
@@ -1,0 +1,78 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+
+namespace Mapster.Tests
+{
+    [TestClass]
+    public class WhenMappingNullableAbstractWithInclude
+    {
+        [TestMethod]
+        public void Null_Abstract_Property_With_Include_Maps_To_Null()
+        {
+            var config = new TypeAdapterConfig();
+            config.Default.MapToConstructor(true);
+            config
+                .NewConfig<DtoBase, DomainBase>()
+                .Include<DtoDerived, DomainDerived>();
+
+            var dto = new DtoFoo { Field = null };
+
+            var domain = dto.Adapt<DomainFoo>(config);
+
+            domain.Field.ShouldBeNull();
+        }
+
+        [TestMethod]
+        public void Derived_Abstract_Property_With_Include_Maps_To_Derived_Domain()
+        {
+            var config = new TypeAdapterConfig();
+            config.Default.MapToConstructor(true);
+            config
+                .NewConfig<DtoBase, DomainBase>()
+                .Include<DtoDerived, DomainDerived>();
+
+            var dto = new DtoFoo
+            {
+                Field = new DtoDerived { Id = "id" },
+            };
+
+            var domain = dto.Adapt<DomainFoo>(config);
+
+            domain.Field.ShouldNotBeNull();
+            domain.Field.ShouldBeOfType<DomainDerived>();
+            ((DomainDerived)domain.Field).Id.ShouldBe("id");
+        }
+
+        #region test classes
+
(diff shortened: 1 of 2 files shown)


wyf027:

wyf027:

Issue #925, opened 2026-04-22
TypeAdapterConfig.GlobalSettings.Compile() throws exception when open generics are configured


cjung95:
When a mapping for a generic type is configured like described [here ](https://github.com/chaowlert/Mapster/wiki/Configuration#open-generics) `TypeAdapterConfig.GlobalSettings.Compile()` throws a Mapster.CompileException: 'Operation is not valid due to the current state of the object.' with an inner ArgumentException: Type ClassA`1[T] is a generic type definition (Parameter 'type').

Minimal working example:
```c#
using Mapster;

TypeAdapterConfig.GlobalSettings.ForType(typeof(ClassA<>), typeof(ClassB<>));
TypeAdapterConfig.GlobalSettings.Compile();

var classA = new ClassA<int> { Variable = 15 };
var classB = classA.Adapt<ClassB<int>>();

class ClassA<T> { public T? Variable { get; set; } = default; }
class ClassB<T> { public T? Variable { get; set; } = default; }
``` 

When removing `TypeAdapterConfig.GlobalSettings.Compile();` the mapping works just fine.

The exception is thrown in `Mapster.Adapters.BaseAdapter.CreateAdaptFunc(CompileArgument arg)` when calling `Expression.Parameter(arg.SourceType)`.

Mapster version 10.0.7

DocSvartz:
Hi @cjung95  I checked, this isn't a regression in v10. It also threw an error in version 7.4.0.
It seems to actually work like a hidden `Inherits` - when a child class copies the parent's settings. But the OpenGeneric configuration itself ends up in the shared settings store.
And since it's impossible to create an OpenGeneric instance, it doesn't compile, as it's impossible to create a function that instantiates OpenGeneric.

cjung95:
Hi @DocSvartz,
thanks for you reply.
I'm running the Compile() function to validate my configuration,  so I need a solution here.
Should I figure out a solution myself and create a pull request, or will someone on your team take on this issue?

DocSvartz:
I'm currently working on a new way to create config for OpenGeneric
using marker

```
TypeAdapterConfig<ClassA<IOpenGeneric>, ClassB<IOpenGeneric>>
.NewConfig()
.Map(dest => dest.Variable2, src => src.Variable);
```

This problem won't occur there.

In your case, can remove the error by removing the check for Opengeneric types.
Because the check result is actually valid. 😅

cjung95:
[...]

PR #965, opened 2026-05-25
test: ensure Compile skips open generic mapping rules (#925)

wyf027:
## Summary
- Add regression coverage for configuring `ClassA<>` → `ClassB<>` and calling `Compile()` without throwing
- Verifies closed generic mappings still work after validation

Addresses #925

## Test plan
- [ ] CI passes (`WhenMappingWithOpenGenerics.Compile_With_Open_Generic_Mapping_Does_Not_Throw`)


src/Mapster.Tests/WhenMappingWithOpenGenerics.cs
DocSvartz: @leno23  If you added a separate test, then delete it. And everything will be fine. 👍

DocSvartz (approved):

Issue #947, opened 2026-05-15
AllowImplicitDestinationInheritance(true) Combines all settings, including those that are not recommended for use by child type


bill2004158:
Sample Code:

````

    public class Animal { }
    public class Dog : Animal { }
    public class Bird : Animal { }


    public class AnimalDto { }
    public class DogDto : AnimalDto { }
    public class BirdDto : AnimalDto { }

    class Program
    {
        static void Main(string[] args)
        {
            var config = new TypeAdapterConfig();
            config.NewConfig<AnimalDto, Animal>().MapWith((s) => s is BirdDto ? new Bird() : new Dog());
            config.AllowImplicitDestinationInheritance = true;

            var d = new DogDto();
            var a = d.Adapt<Dog>(config);
        }
    }

**Comment by DocSvartz** [comment]
Hi, @bill2004158. This isn't supposed to work that way.
Polymorphism works the other way around: a variable in a child class can be assigned to a parent class.

[MapWith ](https://mapstermapper.github.io/Mapster/articles/settings/Custom-conversion-logic.html?q=MapWith)- This is a custom object transformation function. It replaces Mapster's work.

```
Expression<Func<AnimalDto, Animal>> expressionMapWith = (s) => s is BirdDto ? new Bird() : new Dog(); // result is type Animal 

Dog Dto2 = (Dog)new Animal(); // Unable to cast

Expression<Func<DogDto, Dog>> expression = (s) => s is BirdDto ? new Bird() : new Dog(); // Unable to cast
```

bill2004158:
hi @DocSvartz ,
we have factory to create the instance of Animal.
while in Mapping config, we don't register the concrete class.
thus "AllowImplicitDestinationInheritance = true" is used.

if create config for each concrete class,
then there will be many "NewConfig<,>()".
which will be hard to manage.

DocSvartz:
Does Animal require a special constructor? 
Or do the field names in it and its heirs not match?

bill2004158:
Expression<Func<AnimalDto, Animal>> expressionMapWith = (s) => s is BirdDto ? new Bird() : new Dog(); // result is type Animal 
-->
result is type Animal,
but the real type is Bird or Dog.
which could be cause to Dog in correct condition?

Expression<Func<DogDto, Dog>> expression = (s) => s is BirdDto ? new Bird() : new Dog(); // Unable to cast
-->
correct, but it defined "NewConfig<AnimalDto, Animal>()".
[...]

Issue #943, opened 2026-05-07
EmptyCollectionIfNull no longer initializes empty collections in reccords for nullable source


moikot:
I believe commit 1117602c5b0db272d48dd645acd44c46157c6f56 introduced in [v10.0.7-pre1](https://github.com/MapsterMapper/Mapster/releases/tag/v10.0.7-pre01) a regression in Mapster.

Previously, when using the EmptyCollectionIfNull destination transform, a null collection in the source would correctly map to an empty collection in the destination. After this change, that behavior no longer works.

```csharp
class FooDto
{
    public string[] Strings { get; set; }
}

record Foo(List<string> Strings);

// Configure
TypeAdapterConfig.GlobalSettings.Default
    .AddDestinationTransform(DestinationTransform.EmptyCollectionIfNull);

// Arrange
var fooDto = new FooDto();

// Act
var foo = fooDto.Adapt<Foo>();

// Assert
foo.Strings.ShouldNotBeNull();
```

## Expected Behavior
foo.Strings should be initialized as an empty array (string[0]), not null.

## Actual Behavior
foo.Strings is null.

## Impact
This regression makes migration from AutoMapper significantly more difficult, as AutoMapper applies this behavior by default (null collections are mapped to empty ones).

DocSvartz:
Hi, @moikot based on origin tests, this settings was intended to Properties or members of destination Type that are of collection.
So, in terms of constructor parameters, it was undocumented behavior.

I have a couple of ideas on how to get this back.

bill2004158:
https://github.com/MapsterMapper/Mapster/blob/master/src/Mapster/TypeAdapter.cs#L47
if the source is null, 
it return 'default' value directly, without any "EmptyCollectionIfNull destination transform".
is it correct?

I thought the transform should be applied before return the result?

DocSvartz:
> if the source is null,
> it return 'default' value directly, without any "EmptyCollectionIfNull destination transform".
> is it correct?

Yes, It is behavior by default for all type. 

As I said before, `EmptyCollectionIfNull `was intended only for Property or Field **participating in mapping**. 
That is, if the source does not have a Property or Field that can be mapped to the Property or Field in Destination type, this will also not work. 
Mapster only transfers data from the Source to the Destination. It does not modify the Type instantiation process.


```
 [TestMethod]
 public void DestinationTransform943_2()
 {
      var config = new TypeAdapterConfig();

  config.Default
[...]

Issue #938, opened 2026-04-28
Does Mapster has a way to explicitly map the same object while false ShallowCopyForSameType?


klabajo:
## Is this possible?
```cs
cfg.Default.ShallowCopyForSameType(true);

// Attempted code:
cfg.NewConfig<MyStuff, MyStuff>().ShallowCopyForSameType(false);
```

## Trying to replicate similar behavior in the following:

> // Source - https://stackoverflow.com/a/31677062
// Posted by Jimmy Bogard
// Retrieved 2026-04-28, License - CC BY-SA 3.0
> You're reusing a type on both the source and destination objects, "MyStuff". When AutoMapper sees two assignable types, it assigns them rather than copying them. You can override this behavior by creating an explicit map:
```cs
Mapper.CreateMap<MyStuff, MyStuff>();
```
> AutoMapper defaults to assigning, as AutoMapper is not a copying/cloning library.

## Concerns
- While I'm trying out attempted code, some passes and some fails the unit test wherein the failed attempted code always had an error `System.NullReferenceException : Object reference not set to an instance of an object.`. 
- These are only occurring for objects that has their own parameterless constructor that instantiates their uninitialized object properties if it's still null.
```cs
public FailingMyStuff() { CreateEmptyEntities(); }
public void CreateEmptyEntities() { Item1 ??= new(); Item2 ??= new(); }
```
- Although it can be fixed using `MapWith` and defining a function, let's say `MapToSelf`, where it's general structure is `return src == null ? null : new MyStuff(){ ... };` and I've verified that this passed the failing unit tests.

### However, I wanted to know if there's a better way in automating this instead of using `MapToSelf`.

DocSvartz:
HI, @klabajo. What version of mapster are you using? Try latest pre-release version.

You do not have to register custom mapping configuration If you want to receive a shallow copy.

`cfg.NewConfig<MyStuff, MyStuff>()` ❌

klabajo:
I'm on version 10.0.7, I'll try out pre-release

klabajo:
@DocSvartz Oh also, I have `cfg.RequireExplicitMapping = true;` forgot to mention. So I still need to register the config `cfg.NewConfig<MyStuff, MyStuff>()`

klabajo:
It didn't seem to work while using pre-release version, still the same error as `System.NullReferenceException : Object reference not set to an instance of an object.`. 

[...]

Issue #900, opened 2026-03-27
Mapping an object with getter only properties seems to be broken in v10.0.6 (works as expected in 7.4.0)


ripebananas:
Hello. I am trying to map an object with getter-only properties using a custom `Map` functon. This seems to be broken in v10.0.6, while it was fine in v7.4.0. See the following sample:
```c#
var typeAdapterConfig = new Mapster.TypeAdapterConfig();
typeAdapterConfig.Scan(typeof(Program).Assembly);
typeAdapterConfig.Compile();
var mapper = new MapsterMapper.Mapper(typeAdapterConfig);

var dto = new Dto { Data = new("new") };
Console.WriteLine(mapper.Map(dto, new Entity { Data = new("old") }).Data?.Value); // outputs "old" with v10.0.6, outputs "new" with v7.4.0
Console.WriteLine(mapper.Map(dto, new Entity()).Data?.Value); // outputs "new" with both v10.0.6 and v7.4.0
Console.WriteLine(mapper.Map<Entity>(dto).Data?.Value); // outputs "new" with both v10.0.6 and v7.4.0

class Dto { public Data? Data { get; set; } }
class Entity { public Data? Data { get; set; } }
class Data(string value) { public string Value => value; }

class Mappings : Mapster.IRegister
{
    public void Register(Mapster.TypeAdapterConfig config)
    {
        config.NewConfig<Dto, Entity>()
            .Map(x => x.Data, x => x.Data == null ? null : new Data(x.Data.Value));
    }
}
```

Specifically this case seems to be broken:
```c#
var dto = new Dto { Data = new("new") };
var entity = mapper.Map(dto, new Entity { Data = new("old") });
```
I would expect the result entity's `Data` property to be mapped correctly from the source dto, as specified in the mapping function:
`config.NewConfig<Dto, Entity>().Map(x => x.Data, x => x.Data == null ? null : new Data(x.Data.Value));`
As noted, Mapster v7.4.0 seems to be find, but v10.0.* seems to be broken.
Or maybe I am missing something here, please advise, thank you.

DocSvartz:
Hi @ripebananas 
Due to numerous issues, classes with this signature 
`class Data(string value) { public string Value => value; }`
are no longer automatically recognized as [record types](https://mapstermapper.github.io/Mapster/articles/mapping/Data-types.html?tabs=Records-v7-4-0#record-types). and [release notes](https://github.com/MapsterMapper/Mapster/releases)

[...]

PR #966, opened 2026-05-25
fix: skip destination transforms for explicit member maps (#952, ctor path)

### LINKED ISSUE #952: DestinationTransform.EmptyCollectionIfNull overrides explicit null mapping

wyf027:
## Summary
- Explicit `.Map(...)` configuration now takes precedence over global `DestinationTransform` rules such as `EmptyCollectionIfNull`
- Prevents explicit null member mappings from being overridden by collection transforms

Fixes #952

## Test plan
- [ ] CI passes (`WhenPerformingDestinationTransforms.Explicit_Null_Mapping_Is_Not_Overridden_By_EmptyCollectionIfNull`)

### DIFF
--- a/src/Mapster.Tests/WhenPerformingDestinationTransforms.cs
+++ b/src/Mapster.Tests/WhenPerformingDestinationTransforms.cs
@@ -95,6 +95,19 @@ namespace Mapster.Tests
             destination.Set.Count.ShouldBe(0);
         }
 
+        [TestMethod]
+        public void Explicit_Null_Mapping_Is_Not_Overridden_By_EmptyCollectionIfNull()
+        {
+            var config = new TypeAdapterConfig();
+            config.Default.AddDestinationTransform(DestinationTransform.EmptyCollectionIfNull);
+            config.NewConfig<ExplicitNullSource, ExplicitNullDestination>()
+                .Map(d => d.Strings, _ => (string[]?)null);
+
+            var destination = new ExplicitNullSource([]).Adapt<ExplicitNullDestination>(config);
+
+            destination.Strings.ShouldBeNull();
+        }
+
         #region TestClasses
 
         public class SimplePoco
@@ -143,6 +156,13 @@ namespace Mapster.Tests
             public double[,] MultiDimentionalArray { get; set; }
             public IReadOnlyDictionary<string, ChildDto> ChildDict { get; set; }
             public ISet<string> Set { get; set; }
+        }
+
+        public record ExplicitNullSource(string?[] Strings);
+
+        public class ExplicitNullDestination
+        {
+            public string[]? Strings { get; set; }
         }
 
         #endregion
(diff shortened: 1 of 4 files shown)


wyf027:
Updated `8b8898a`: merged ctor-path fix from former #973 — `ApplyDestinationTransform` on record/class ctor args + explicit `.Map` skip in `ApplyNullPropagationFromCtor`. Added `Explicit_Null_Record_Ctor_Mapping_Is_Not_Overridden_By_EmptyCollectionIfNull` alongside existing property-map regression (#952). Covers both adapter and ctor instantiation paths.

wyf027:
**CI note:** latest push `8b8898a` — workflow still **action_required** (fork). Once **Build and Test** is approved, this covers both paths:\n- property mapping (`BaseAdapter`)\n- record/class ctor args (`BaseClassAdapter` + `ApplyDestinationTransform`)\n\nBuilds on #945 (#943) already on `development`.

wyf027:
Batch ping: **#966**, **#970**, **#971**, **#976**, **#979** (and other open fork PRs) are rebased on latest `development` (`11853b3`) and mergeable — all waiting on maintainer **Build and Test** workflow approval. Happy to address review feedback once CI runs.

wyf027:
**Open fork PRs (2026-05-27)** — all rebased on `development` (`11853b3`), mergeable, awaiting **Build and Test** workflow approval:

| PR | Type |
|----|------|
| #979 | test (#938 UseDestinationValue) |
| #966 | fix (#952 + ctor) |
| #970 | fix (#900 MapToTarget) |
| #971 | fix (#898 ProjectToType) |
| #976 | fix (#947 MapWith cast) |
| #977 | fix (#801 projection Include) |
| #965, #972 | test-only |

Happy to address review feedback once CI runs.

PR #968, opened 2026-05-25
fix: use type default for optional DateTime constructor params (#954)

### LINKED ISSUE #954: CompileException when mapping record with DateTime to class with optional constructor parameter

wyf027:
## Summary
- Extract `CreateOptionalParameterDefault` for optional constructor parameters.
- When `DefaultValue` is null or `DBNull`, emit `destinationType.CreateDefault()` instead of `Expression.Constant(null, valueType)`.
- Adds regression test mapping a record with `DateTime` to a class with an optional `DateTime` constructor parameter.

Fixes #954

## Test plan
- [x] Added `WhenMappingRecordWithOptionalDateTimeConstructor` regression test

### DIFF
--- a/src/Mapster/Adapters/BaseClassAdapter.cs
+++ b/src/Mapster/Adapters/BaseClassAdapter.cs
@@ -225,23 +225,9 @@ namespace Mapster.Adapters
                 Expression defaultConst;
                 Expression getter;
 
-#if NETSTANDARD2_0
-                try
-                {
-                    defaultConst = parameterInfo.IsOptional && parameterInfo.DefaultValue != null
-                    ? Expression.Constant(parameterInfo.DefaultValue, member.DestinationMember.Type)
+                defaultConst = parameterInfo.IsOptional
+                    ? CreateOptionalParameterDefault(parameterInfo, member.DestinationMember.Type)
                     : parameterInfo.ParameterType.CreateDefault();
-                }
-                catch (FormatException)
-                {
-                    defaultConst = parameterInfo.ParameterType.CreateDefault();
-                }
-
-#else
-                defaultConst = parameterInfo.IsOptional && parameterInfo.DefaultValue != null
-                    ? Expression.Constant(parameterInfo.DefaultValue, member.DestinationMember.Type)
-                    : parameterInfo.ParameterType.CreateDefault();
-#endif
                 
                 if (member.Getter == null)
                 {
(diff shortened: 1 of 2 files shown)


DocSvartz:
Hi, @leno23  Thanks, but it is fixed.  and in net48 and perhaps in netcore version 6 or below, instead of the default value being null, the default value actually contains a FormatException.

src/Mapster/Adapters/BaseClassAdapter.cs
```diff
@@ -289,6 +275,15 @@ protected Expression CreateInstantiationExpression(Expression source, ClassMappi
             return Expression.New(classConverter.ConstructorInfo!, arguments);
         }
 
+        static Expression CreateOptionalParameterDefault(ParameterInfo parameterInfo, Type destinationType)
+        {
+            var defaultValue = parameterInfo.DefaultValue;
+            if (defaultValue == null || defaultValue is DBNull)
```
DocSvartz, line 281: @leno23  `defaultValue is DBNull` Please add a test for this case if it applies. There is already a test for the general case.

wyf027:
Thanks @DocSvartz — confirmed the `FormatException` guard for optional `DateTime` ctor params is already on `development` (`BaseClassAdapter`, `#if NETSTANDARD2_0`). Closing as duplicate of the upstream fix.

Appreciate the note about net48 / net6-or-below default-value behavior vs null.

wyf027:
Superseded by fix already on development branch.

PR #970, opened 2026-05-25
fix: apply custom Map results when mapping to existing target (#900)

### LINKED ISSUE #900: Mapping an object with getter only properties seems to be broken in v10.0.6 (works as expected in 7.4.0)

wyf027:
## Summary
- Fixes `MapToTarget` (`mapper.Map(dto, existingEntity)`) ignoring custom `.Map()` resolvers when the destination member already has a value.
- When a destination member has an explicit custom `.Map()` configuration, skip passing the existing destination member into nested adaptation so the resolver result is assigned directly.
- Adds regression tests reproducing issue #900 for MapToTarget with pre-existing destination values, null destination members, and new-object mapping.

## Root cause
For `MapToTarget`, `ClassAdapter` always passed the existing destination member expression into `CreateAdaptExpression`. That triggered nested `MapToTarget` adaptation for the member type. For immutable/getter-only types (like primary-constructor classes), nested mapping preserves the existing destination instance instead of replacing it with the custom `.Map()` result.

## Test plan
- [ ] CI passes on `Mapster.Tests`
- [ ] `WhenMappingGetterOnlyMapToTargetRegression.MapToTarget_WithCustomMap_ShouldReplaceExistingDestinationMemberValue`
- [ ] Existing MapToTarget regression tests remain green

Fixes #900

### DIFF
--- a/src/Mapster.Tests/WhenMappingGetterOnlyMapToTargetRegression.cs
+++ b/src/Mapster.Tests/WhenMappingGetterOnlyMapToTargetRegression.cs
@@ -1,0 +1,75 @@
+using MapsterMapper;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+
+namespace Mapster.Tests;
+
+/// <summary>
+/// https://github.com/MapsterMapper/Mapster/issues/900
+/// </summary>
+[TestClass]
+public class WhenMappingGetterOnlyMapToTargetRegression
+{
+    [TestMethod]
+    public void MapToTarget_WithCustomMap_ShouldReplaceExistingDestinationMemberValue()
+    {
+        var config = new TypeAdapterConfig();
+        config.NewConfig<Dto900, Entity900>()
+            .Map(x => x.Data, x => x.Data == null ? null : new Data900(x.Data.Value));
+
+        var mapper = new Mapper(config);
+
+        var dto = new Dto900 { Data = new Data900("new") };
+        var entity = mapper.Map(dto, new Entity900 { Data = new Data900("old") });
+
+        entity.Data.ShouldNotBeNull();
+        entity.Data!.Value.ShouldBe("new");
+    }
+
+    [TestMethod]
+    public void MapToTarget_WithCustomMap_ShouldStillWorkWhenDestinationMemberIsNull()
+    {
+        var config = new TypeAdapterConfig();
+        config.NewConfig<Dto900, Entity900>()
+            .Map(x => x.Data, x => x.Data == null ? null : new Data900(x.Data.Value));
+
+        var mapper = new Mapper(config);
+
+        var dto = new Dto900 { Data = new Data900("new") };
+        var entity = mapper.Map(dto, new Entity900());
+
+        entity.Data.ShouldNotBeNull();
+        entity.Data!.Value.ShouldBe("new");
+    }
+
+    [TestMethod]
+    public void Map_WithCustomMap_ShouldStillWorkForNewDestination()
+    {
+        var config = new TypeAdapterConfig();
+        config.NewConfig<Dto900, Entity900>()
+            .Map(x => x.Data, x => x.Data == null ? null : new Data900(x.Data.Value));
+
+        var mapper = new Mapper(config);
+
+        var dto = new Dto900 { Data = new Data900("new") };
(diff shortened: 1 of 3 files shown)


src/Mapster.Tests/WhenMappingGetterOnlyMapToTargetRegression.cs
```diff
@@ -0,0 +1,75 @@
...
+        entity.Data.ShouldNotBeNull();
+        entity.Data!.Value.ShouldBe("new");
+    }
+
+    [TestMethod]
+    public void MapToTarget_WithCustomMap_ShouldStillWorkWhenDestinationMemberIsNull()
+    {
+        var config = new TypeAdapterConfig();
+        config.NewConfig<Dto900, Entity900>()
+            .Map(x => x.Data, x => x.Data == null ? null : new Data900(x.Data.Value));
```
DocSvartz, line 34: @andrerav  @stagep  
I think users want something like this.

```
 config.NewConfig<Poco, Dto>()
     .Map(dto => dto.Age, src => src.AgeString)
     .OvverideTypesConfigForMapPair(cfg=> 
     cfg.DefaultValue("42"));
```

But Map doesn't override the type mapping behavior; it can simply [provide another alternative source](https://mapstermapper.github.io/Mapster/articles/settings/custom/Custom-mapping.html?q=Map%20condition#mapping-with-condition).

Example For this case:

```
config.NewConfig<Dto900, Entity900>()
            .Map(x => x.Data, x => x.Data ) 
.OvverideTypesConfigForMapPair(cfg=> 
     cfg.MapWith(x => x.Data == null ? null : new Data900(x.Data.Value));

```

wyf027:
Fixes #900 — custom `.Map(...)` results now apply on `MapToTarget` / getter-only destinations (`ClassAdapter` + `BaseClassAdapter`). Branch rebased on latest `development`. Awaiting workflow approval for CI.

wyf027:
Thanks @DocSvartz — `OverrideTypesConfigForMapPair` (or similar) would be a cleaner long-term API for overriding nested type-mapping behavior from a member `.Map`.

This PR stays intentionally minimal: **#900** reports that an explicit `.Map(x => x.Data, …)` is ignored on **MapToTarget** because `ClassAdapter` always prefers `UseDestinationValue`/destination member when `MapType.MapToTarget`, even when a custom resolver exists. The fix only skips destination-member reuse when a custom map resolver is present (`HasCustomMemberMap`).

Happy to pivot to a config-override API if maintainers prefer that direction — the regression tests should still apply either way.

Issue #898, opened 2026-03-26
ProjectToType throws when nullable navigation maps to record DTO with non-nullable value type


unzeitip:
First of all, thank you for Mapster — it's been a great fit for our project and we really appreciate the work you've put into it.

  I've run into what looks like a bug with `ProjectToType` when mapping a nullable navigation property to a record DTO that has non-nullable value type parameters. Here I provide additional context if it helps track this down.

  **Mapster version:** 10.0.6
  **EF Core version:** 10.0.5

  ### Minimal reproduce

  ```csharp
// Entities:
  public class OrderEntity
  {
      public int Id { get; set; }
      public int? CodId { get; set; }
      public OrderCodEntity? Cod { get; set; }
  }

  public class OrderCodEntity
  {
      public int Id { get; set; }
      public long Value { get; set; }
  }

  // DTOs:
  public record OrderDto(int Id, OrderCodDto? Cod);
  public record OrderCodDto(long Value);

  // Mapping + query:
  TypeAdapterConfig<OrderEntity, OrderDto>.NewConfig();
  TypeAdapterConfig<OrderCodEntity, OrderCodDto>.NewConfig();

  var results = dbContext.Orders
      .ProjectToType<OrderDto>()
      .ToList();  // throws when any Order has Cod == null

  // Seed data:
  new OrderEntity { Id = 1, Cod = new OrderCodEntity { Value = 42L } }  // OK
  new OrderEntity { Id = 2, Cod = null }                               // causes crash
```

  #### Expected behavior

  OrderDto(1, OrderCodDto(42))
  OrderDto(2, null)

 #### Actual behavior

  System.InvalidOperationException: Nullable object must have a value.

 #### Workaround

   ```csharp TypeAdapterConfig<OrderEntity, OrderDto>.NewConfig()
      .Map(
          d => d.Cod,
          s => s.Cod == null ? null : new OrderCodDto(s.Cod.Value)
      );
```

  #### Note

  The issue persists when using `Mapster.EFCore` 10.0.6 and calling `EFCoreProjectToType<OrderDto>()` instead of `ProjectToType<OrderDto>()` - same exception is thrown.

DocSvartz:
Hi @unzeitip. 

Does this error exist when using?

Mapster version: 10.0.4-a
EF Core version: 10.0.0

DocSvartz:
@unzeitip It's weird, but in your example I get a different error: System.NullReferenceException: Object reference not set to an instance of an object.
[...]

PR #971, opened 2026-05-25
fix: ProjectToType nullable navigation to record DTO ctor (#898)

wyf027:
## Summary

- Fixes `ProjectToType` throwing `InvalidOperationException: Nullable object must have a value` when a nullable navigation (e.g. `Cod == null`) maps to a record DTO constructor parameter whose nested record has non-nullable value-type parameters.
- Adds an explicit null guard for nullable getters mapping to record constructor parameters (same pattern as abstract/non-public ctor types).
- Tracks `NullChecks` by expression identity instead of nullable type so nested projection parameters are not incorrectly skipped.
- Strengthens the #898 regression test with compile + execution assertions.

## Root cause

Constructor mapping for record DTOs could still evaluate nested record constructor arguments (e.g. `Cod.Value`) when the nullable navigation was null. PR #903 improved ctor null propagation, but nullable navigation → nullable record ctor parameters still needed an explicit short-circuit before nested record mapping.

## Test plan

- [x] Added/updated regression test `WhenAddCtorNullablePropagation.NullablePropagationFromCtorWorking`
- [ ] CI green on `Mapster.Tests`

Fixes MapsterMapper/Mapster#898

### DIFF
--- a/src/Mapster.Tests/WhenAddCtorNullablePropagation.cs
+++ b/src/Mapster.Tests/WhenAddCtorNullablePropagation.cs
@@ -1,5 +1,5 @@
-﻿using Mapster.Tests.Classes;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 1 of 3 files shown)


wyf027:
Fixes #898 — moves ctor `NullChecks` registration out of per-member loop and tightens `ApplyNullPropagationFromCtor` for `ProjectToType` nullable navigation → record DTO. Builds on #903 (`development`). Test assertions expanded in `WhenAddCtorNullablePropagation`.

Issue #918, opened 2026-04-17
[MapToTarget] [10.0.7] Mutable type not mutate when SourceType == DestinationType


Ultramund1:
我有一个类如下：
`using CommunityToolkit.Mvvm.ComponentModel;`
using SignalSet.Utilities;
using System.ComponentModel;
using System.ComponentModel.DataAnnotations;
using System.Net;
using System.Text;

namespace SignalSet.Models;
```

public sealed class Station : ObservableValidator
{
    public int RowNo
    {
        get;
        set => SetProperty(ref field, value);
    }

    public string Location
    {
        get;
        set => SetProperty(ref field, value);
    } = string.Empty;

    public string RealIp
    {
        get;
        set => SetProperty(ref field, value);
    } = string.Empty;

    [RegularExpression(IpPattern, ErrorMessage = "不是一个有效的IP地址")]
    public string Ip
    {
        get;
        set => SetProperty(ref field, value, true);
    } = string.Empty;

    [DisplayName("端口"), Range(1000, 65535, ErrorMessage = "端口值必须设置为[1000-65535]之间的数字")]
    public int Port
    {
        get;
        set => SetProperty(ref field, value, true);
    }

    [RegularExpression(IpPattern, ErrorMessage = "不是一个有效的IP地址")]
    public string Server
    {
        get;
        set => SetProperty(ref field, value, true);
    } = string.Empty;

    [RegularExpression(IpPattern, ErrorMessage = "不是一个有效的IP地址")]
    public string SubMask
    {
        get;
        set => SetProperty(ref field, value, true);
    } = string.Empty;

    [RegularExpression(IpPattern, ErrorMessage = "不是一个有效的IP地址")]
    public string GateWay
    {
        get;
        set => SetProperty(ref field, value, true);
    } = string.Empty;

    public WorkMode Mode
    {
        get;
        set => SetProperty(ref field, value);
    }

    public int SecurityKey
    {
        get;
        set => SetProperty(ref field, value);
    }

    public string Address
    {
        get;
        set => SetProperty(ref field, value);
    } = string.Empty;

    public string OldName
    {
        get;
        set => SetProperty(ref field, value);
    } = string.Empty;

    public string DbAddress
    {
        get;
        set => SetProperty(ref field, value);
    } = string.Empty;

    [RegularExpression(MacPattern, ErrorMessage = "不是一个有效的MAC地址")]
    public string Mac
    {
        get;
        set => SetProperty(ref field, value, true);
[...]

PR #972, opened 2026-05-25
test: cover MapToTarget for types with read-only base properties (#918)

wyf027:
## Summary
- Adds regression test for same-type `Adapt(target)` when the type inherits read-only base members (ObservableValidator-like pattern) alongside writable properties.

Relates to #918 (fixed on `development` via `IsNotSelfCreation` `Any` → `All` change).

## Test plan
- [x] `MapToTarget_SameType_WithMixedReadOnlyBaseProperties_UpdatesDestination`


wyf027:
Regression test only — #918 fix (`IsNotSelfCreation`) is already on `development` (388602b). Ready for review once CI workflow is approved.

PR #973, opened 2026-05-25
fix: apply EmptyCollectionIfNull to record constructor args (#943)

### LINKED ISSUE #943: EmptyCollectionIfNull no longer initializes empty collections in reccords for nullable source

wyf027:
## Summary
- Apply `EmptyCollectionIfNull` (and other destination transforms) to record/class constructor arguments when null sources would otherwise skip transforms.
- Skip destination transforms when a member has an explicit `.Map()` resolver so explicit null mappings are preserved.

Fixes #943
Relates to #952 (ctor-parameter path; property path covered by #966)

## Test plan
- [x] `NullableCtorPropagationCurrentWorkWithDestinationTransform` (#943)
- [x] `ExplicitNullMapping_ShouldNotBeOverridden_ByDefaultEmptyCollectionTransform` (#952)

### DIFF
--- a/src/Mapster.Tests/WhenExplicitNullDestinationTransformRegression.cs
+++ b/src/Mapster.Tests/WhenExplicitNullDestinationTransformRegression.cs
@@ -1,0 +1,38 @@
+using System.Collections.Generic;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+
+namespace Mapster.Tests
+{
+    /// <summary>
+    /// https://github.com/MapsterMapper/Mapster/issues/952
+    /// </summary>
+    [TestClass]
+    public class WhenExplicitNullDestinationTransformRegression
+    {
+        [TestMethod]
+        public void ExplicitNullMapping_ShouldNotBeOverridden_ByDefaultEmptyCollectionTransform()
+        {
+            var config = new TypeAdapterConfig();
+
+            config.Default
+                .AddDestinationTransform(DestinationTransform.EmptyCollectionIfNull);
+
+            config.NewConfig<Foo952, FooDto952>()
+                .Map(d => d.Strings, _ => (string[]?)null);
+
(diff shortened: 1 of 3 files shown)


wyf027:
Heads-up for review: #943 ctor-path transforms were partially addressed upstream in #945 (`development`). This PR additionally wires `ApplyDestinationTransform` + explicit-map skip (`HasExplicitMemberMap`) on the ctor path — overlapping with #966 (#952) on the adapter path. The added regression test currently references **#952** (explicit null vs `EmptyCollectionIfNull`), same scenario as #966; happy to rebase onto latest `development` and split/rename tests if that helps review.

wyf027:
Consolidated into #966: ctor-path `ApplyDestinationTransform` + explicit-map skip now lives on `fix/explicit-null-map-952-dev` with record-ctor regression coverage in `WhenPerformingDestinationTransforms`. #943 baseline remains on `development` via #945.

Issue #2, opened 2014-10-08
Byte Array Property not being translated correctly.


Garwin4j:
Hello, 

I have a Byte[] that store information of a file. After translating it to another Byte[], file becomes corrupt alot of null characters are left at the end.

PR #974, opened 2026-05-25
fix: honor per-type ShallowCopyForSameType overrides (#938)

### LINKED ISSUE #938: Does Mapster has a way to explicitly map the same object while false ShallowCopyForSameType?

wyf027:
## Summary
- When deciding same-type shallow copy, use `GetMergedSettings` for the current source/destination pair instead of only `arg.Settings` and skipping whenever an explicit rule exists.
- Allows `cfg.Default.ShallowCopyForSameType(true)` with `cfg.NewConfig<T,T>().ShallowCopyForSameType(false)` while preserving nested same-type shallow copy for types without an explicit override.

Fixes #938

## Test plan
- [x] `GlobalShallowCopy_WithPerTypeDeepCopy_ShouldRestoreViaMapToTarget`
- [x] `ParentDeepCopyOverride_ShouldNotDisableImplicitNestedShallowCopyForSameType`

### DIFF
--- a/src/Mapster.Tests/WhenPerTypeShallowCopyOverride.cs
+++ b/src/Mapster.Tests/WhenPerTypeShallowCopyOverride.cs
@@ -1,0 +1,108 @@
+using Mapster.Models;
+using MapsterMapper;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+
+namespace Mapster.Tests
+{
+    /// <summary>
+    /// https://github.com/MapsterMapper/Mapster/issues/938
+    /// </summary>
+    [TestClass]
+    public class WhenPerTypeShallowCopyOverride
+    {
+        [TestMethod]
+        public void GlobalShallowCopy_WithPerTypeDeepCopy_ShouldRestoreViaMapToTarget()
+        {
+            var cfg = new TypeAdapterConfig();
+            cfg.RequireExplicitMapping = true;
+            cfg.Default.ShallowCopyForSameType(true);
+            cfg.Default.AvoidInlineMapping(true);
+
+            cfg.NewConfig<RandomObject2, RandomObject2>();
+            cfg.NewConfig<RandomObject1, RandomObject1>();
+
+            cfg.NewConfig<MyFailStuff, MyFailStuff>()
+                .ShallowCopyForSameType(false);
+
+            cfg.GetMergedSettings(new TypeTuple(typeof(MyFailStuff), typeof(MyFailStuff)), MapType.Map)
+                .ShallowCopyForSameType.ShouldBe(false);
+
+            cfg.Compile();
+
+            var mapper = new Mapper(cfg);
+
+            var dynamicStuff = new MyFailStuff();
+            dynamicStuff.Item1 = new RandomObject1();
+            dynamicStuff.Item1.SampleName = "SN1";
+            dynamicStuff.Item2 = new RandomObject2();
+            dynamicStuff.Item2.SampleNumber = 2;
+
+            var originalStuff = mapper.Map<MyFailStuff, MyFailStuff>(dynamicStuff);
+
+            dynamicStuff.Item1.SampleName = "SN1CHANGED";
+            dynamicStuff.Item2.SampleNumber = 3;
+
+            mapper.Map(originalStuff, dynamicStuff);
+
+            dynamicStuff.Item1.SampleName.ShouldBe("SN1");
+            dynamicStuff.Item2.SampleNumber.ShouldBe(2);
+            ReferenceEquals(dynamicStuff.Item1, originalStuff.Item1).ShouldBeFalse();
+            ReferenceEquals(dynamicStuff.Item2, originalStuff.Item2).ShouldBeFalse();
+        }
+
+        [TestMethod]
+        public void ParentDeepCopyOverride_ShouldNotDisableImplicitNestedShallowCopyForSameType()
+        {
+            var cfg = new TypeAdapterConfig();
+            cfg.Default.ShallowCopyForSameType(true);
+            cfg.NewConfig<Container, Container>()
+                .ShallowCopyForSameType(false);
+
+            cfg.Compile();
+
+            var src = new Container { Child = new NestedChild { Value = 1 } };
+            var dest = src.Adapt<Container>(cfg);
+
+            ReferenceEquals(src.Child, dest.Child).ShouldBeTrue();
+        }
+
+        public class Container
+        {
+            public NestedChild? Child { get; set; }
+        }
+
+        public class NestedChild
+        {
+            public int Value { get; set; }
+        }
+
+        public class MyFailStuff
+        {
+            public MyFailStuff()
+            {
+                CreateEmptyEntities();
+            }
+
+            public void CreateEmptyEntities()
+            {
+                Item1 ??= new RandomObject1();
+                Item2 ??= new RandomObject2();
+            }
+
+            public RandomObject1? Item1 { get; set; }
+
+            public RandomObject2? Item2 { get; set; }
+        }
+
+        public class RandomObject1
+        {
+            public string? SampleName { get; set; }
+        }
+
+        public class RandomObject2
+        {
+            public int SampleNumber { get; set; }
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


wyf027:
Fixes #938 — uses merged settings for same-type shallow copy when `RequireExplicitMapping` is enabled.

GitHub Actions workflows are waiting on maintainer approval for fork PRs (`action_required`). Happy to address review feedback once CI can run.

src/Mapster.Tests/WhenPerTypeShallowCopyOverride.cs
```diff
@@ -0,0 +1,108 @@
...
+namespace Mapster.Tests
+{
+    /// <summary>
+    /// https://github.com/MapsterMapper/Mapster/issues/938
+    /// </summary>
+    [TestClass]
+    public class WhenPerTypeShallowCopyOverride
+    {
+        [TestMethod]
+        public void GlobalShallowCopy_WithPerTypeDeepCopy_ShouldRestoreViaMapToTarget()
```
DocSvartz, line 15: This test is not valid and fail ❌

To obtain the desired result, the issue author used an incorrect configuration.

This is what the author wanted to achieve.

```
[TestMethod]
public void GlobalShallowCopy_WithPerTypeDeepCopy_ShouldRestoreViaMapToTarget()
{
    var cfg = new TypeAdapterConfig();
    cfg.RequireExplicitMapping = true;
    cfg.Default.AvoidInlineMapping(true);

    cfg.NewConfig<RandomObject2, RandomObject2>();
    cfg.NewConfig<RandomObject1, RandomObject1>()
        .Include<RandomObject2, RandomObject2>();

    cfg.NewConfig<MyFailStuff, MyFailStuff>()
       // .ShallowCopyForSameType(true)
        .UseDestinationValue(x => x.Item1)
        .UseDestinationValue(x => x.Item2);



    cfg.Compile();

    var mapper = new Mapper(cfg);


   

    var dynamicStuff = new MyFailStuff();
    dynamicStuff.Item1 = new RandomObject1();
    dynamicStuff.Item1.SampleName = "SN1";
    dynamicStuff.Item2 = new RandomObject2() { SampleNumber = 2 };
    //dynamicStuff.Item2.SampleNumber = 2;

    var str = dynamicStuff.BuildAdapter(cfg).CreateMapToTargetExpression<MyFailStuff>();



    var originalStuff = mapper.Map<MyFailStuff, MyFailStuff>(dynamicStuff);

    dynamicStuff.Item1.SampleName = "SN1CHANGED";
    (dynamicStuff.Item2 as RandomObject2).SampleNumber = 3;

    mapper.Map(originalStuff, dynamicStuff);

    dynamicStuff.Item1.SampleName.ShouldBe("SN1");
    (dynamicStuff.Item2 as RandomObject2).SampleNumber.ShouldBe(2);
    ReferenceEquals(dynamicStuff.Item1, originalStuff.Item1).ShouldBeFalse();
    ReferenceEquals(dynamicStuff.Item2, originalStuff.Item2).ShouldBeFalse();
}

  public class RandomObject1
  {
      public string? SampleName { get; set; }
  }

  public class RandomObject2 : RandomObject1
  {
      public int SampleNumber { get; set; }
  }

```

src/Mapster/Adapters/BaseAdapter.cs
```diff
@@ -503,8 +503,12 @@ internal Expression CreateAdaptExpression(Expression source, Type destinationTyp
...
 
-            if (_source.Type == destinationType && arg.Settings.ShallowCopyForSameType == true
-                && notUsingDestinationValue && rule == null)
+            var shallowCopySettings = _source.Type == destinationType
+                ? arg.Context.Config.GetMergedSettings(tuple, arg.MapType)
+                : arg.Settings;
+
+            if (_source.Type == destinationType && shallowCopySettings.ShallowCopyForSameType == true
+                && notUsingDestinationValue)
                 exp = _source;
```
DocSvartz, line 517: This is the wrong level of mapping.
This is where properties are mapped, not types.

> A shallow copy creates a new top-level object but does not duplicate nested elements. Instead, it copies the memory references of nested objects and arrays. This means modifying a nested property in the copy will inadvertently change it in the original object.

If you consider this a new TopLevel, a new object must be created here in any case, and its properties must be copied by reference.
Or if this is a parsing of the properties of the original TopLevel (the one specified in arg).
Then searching for this setting is pointless, since all TopLevel properties must be copied by reference in any case.

wyf027:
**CI fix pushed** (`44b1479`)

`GlobalShallowCopy_WithPerTypeDeepCopy_ShouldRestoreViaMapToTarget` failed because removing `rule == null` caused explicit configs (`RandomObject1`, `RandomObject2`) to shallow-copy nested references under the global default, so `MapToTarget` never restored `Item1`/`Item2` property values.

Restored `&& rule == null` while keeping `GetMergedSettings` for per-type overrides. Implicit nested same-type mappings (e.g. `NestedChild` in the passing test) still shallow-copy; explicit mapped types still deep-map members on `MapToTarget`.

Please re-run / approve CI when convenient.

wyf027:
Friendly ping: CI fix is on `44b1479`. Fork workflow runs need maintainer approval — could someone approve/re-run **Build and Test** when convenient? 🙏

wyf027:
**CI fix #2** (`669a53d`)

Previous fix restored `rule == null` (MapToTarget test ✅) but broke `NewInstanceConfigurationTest`: always using `GetMergedSettings` ignored `ShallowCopyForSameType(true)` inherited from parent cross-type config (D→E).

Now:
- Explicit same-type rule → `GetMergedSettings` (per-type override, #938)
- Inherited shallow-copy enabled → keep `arg.Settings`
- Otherwise → `GetMergedSettings` (global default for nested implicit same-type)

CI re-running.

wyf027:
CI on `669a53d` is still **action_required** (fork workflow approval). Latest commit preserves inherited `ShallowCopyForSameType` on nested same-type maps while honoring explicit per-type overrides via `GetMergedSettings` — distinct from the #941 baseline on `development`. Ready for review once **Build and Test** is approved.

wyf027:
**CI:** `669a53d` still **action_required** (fork workflow). Distinct from merged #941 — enables `Default.ShallowCopyForSameType(true)` with per-type `NewConfig\u003cT,T\u003e().ShallowCopyForSameType(false)` via `GetMergedSettings`, while preserving inherited shallow copy on nested implicit same-type maps.

wyf027:
@DocSvartz Thanks for the review — agreed on both points.

1. **Test**: The failing scenario should use `UseDestinationValue` on nested properties (your snippet), not `ShallowCopyForSameType(false)` on the top-level config.
2. **Code level**: Per-type `ShallowCopyForSameType` via `GetMergedSettings` inside `CreateAdaptExpression` is the wrong layer — shallow copy applies to top-level type mapping, not individual property adaptation.

Closing this PR. #938 is already covered upstream by **#941** (`DirectAssignmentForSameType`). Appreciate the pointer to the correct configuration pattern.

wyf027:
Follow-up: opened **#979** (test-only) with the `UseDestinationValue` configuration from your review snippet, as regression documentation for #938.

PR #975, opened 2026-05-25
fix: honor per-type ShallowCopyForSameType override (#938)

### LINKED ISSUE #938: Does Mapster has a way to explicitly map the same object while false ShallowCopyForSameType?

wyf027:
## Summary
- Fixes #938 by resolving `ShallowCopyForSameType` from merged per-type settings when adapting same-type members, instead of always using the parent mapping's settings.
- A parent self-mapping with `ShallowCopyForSameType(false)` no longer suppresses global shallow-copy behavior for nested implicit same-type members.
- Explicit per-type `ShallowCopyForSameType(false)` overrides still force deep copy for self-mappings, including MapToTarget restore scenarios with ctor-initialized nested properties.

## Root cause
`CreateAdaptExpression` checked `arg.Settings.ShallowCopyForSameType` from the enclosing mapping context. When a parent type opted into deep copy, nested same-type members incorrectly inherited that setting even when their own merged settings/default should remain shallow.

## Test plan
- [x] `WhenPerTypeShallowCopyOverride.GlobalShallowCopy_WithPerTypeDeepCopy_ShouldRestoreViaMapToTarget`
- [x] `WhenPerTypeShallowCopyOverride.ParentDeepCopyOverride_ShouldNotDisableImplicitNestedShallowCopyForSameType`
- [x] Existing `WhenConfiguringMapping` shallow-copy tests pass

### DIFF
--- a/src/Mapster.Tests/WhenPerTypeShallowCopyOverride.cs
+++ b/src/Mapster.Tests/WhenPerTypeShallowCopyOverride.cs
@@ -1,0 +1,105 @@
+using Mapster.Models;
+using MapsterMapper;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+
+namespace Mapster.Tests
+{
+    [TestClass]
+    public class WhenPerTypeShallowCopyOverride
+    {
+        [TestMethod]
+        public void GlobalShallowCopy_WithPerTypeDeepCopy_ShouldRestoreViaMapToTarget()
+        {
+            var cfg = new TypeAdapterConfig();
+            cfg.RequireExplicitMapping = true;
+            cfg.Default.ShallowCopyForSameType(true);
+            cfg.Default.AvoidInlineMapping(true);
+
+            cfg.NewConfig<RandomObject2, RandomObject2>();
+            cfg.NewConfig<RandomObject1, RandomObject1>();
+
+            cfg.NewConfig<MyFailStuff, MyFailStuff>()
+                .ShallowCopyForSameType(false);
+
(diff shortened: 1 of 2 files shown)


wyf027:
Closing as duplicate of #974 — both target #938; keeping #974 which uses merged settings directly for same-type shallow copy and includes the issue reference in tests.

PR #1008, opened 2026-08-19
Fix/1007 ignoreif record property

alihut:
fix: apply IgnoreIf to record properties outside the primary constructor


DocSvartz:
@alihut  Thanks!

PR #1010, opened 2026-08-20
Fix #399 - Fix mapper interface generation for internal types

### LINKED ISSUE #399: mapster tool generates non-public interface method/property implementation
When generating mapper from interface containing mapping method/property with `internal` classes mapster tool generates `internal` implementation of this method/property

- Mapster.Core v1.2.0
- mapster.tool v8.2.1
- executed on `net5`

```
[Mapper(IsInternal = true)]
internal interface IMyMappers
{
    PocoDto Map(Poco source);

    Expression<Func<Poco, PocoDto>> ProjectToDto { get; }
}

public record Poco(string Name, string Description);

internal record PocoDto(string Name, string Description);
```

Generated mapper class is:
```
internal partial class MyMappers : IMyMappers
{
    internal Expression<Func<Poco, PocoDto>> ProjectToDto => p1 => new PocoDto(p1.Name, p1.Description);
    internal PocoDto Map(Poco p2)
    {
        return p2 == null ? null : new PocoDto(p2.Name, p2.Description);
    }
}
```
> CS0737: 'MyMappers.Map(Poco)' cannot implement an interface member because it is not public.
> CS0737: 'MyMappers.ProjectToDto' cannot implement an interface member because it is not public.

Maybe `internal` class with proper implementations should be generated even in case `MapperAttribute.IsInternal` hasn't been specified

andrerav:
Fixes mapper generation for interfaces that use internal source or destination types. The generated mapper class now uses the same effective visibility rule for both the class modifier and generated mapper members. If the generated class must be internal, its implicit interface implementation members can still be emitted as public.

Fixes #399.

### DIFF
--- a/src/ExpressionTranslator/ExpressionTranslator.cs
+++ b/src/ExpressionTranslator/ExpressionTranslator.cs
@@ -56,6 +56,15 @@ namespace ExpressionDebugger
                 _indentLevel++;
         }
 
+        private bool IsGeneratedTypeInternal()
+        {
+            if (Definitions == null)
+                return false;
+
+            return Definitions.IsInternal ||
+                   Definitions.Implements?.Any(it => !it.GetTypeInfo().IsVisible) == true;
+        }
+
         public static ExpressionTranslator Create(Expression node, ExpressionDefinitions? definitions = null)
         {
             var translator = new ExpressionTranslator(definitions);
@@ -1201,7 +1210,7 @@ namespace ExpressionDebugger
                 if (type == LambdaType.PublicLambda)
                 {
                     var name = methodName ?? "Main";
-                    if (!isInternal)
+                    if (!isInternal && !IsGeneratedTypeInternal())
                         isInternal = node.ReturnType.GetTypeInfo().IsNotPublic ||
                                      node.Parameters.Any(it => it.Type.GetTypeInfo().IsNotPublic);
                     WriteModifierNextLine(isInternal ? "internal" : "public");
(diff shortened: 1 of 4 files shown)


src/ExpressionTranslator/ExpressionTranslator.cs
```diff
@@ -56,6 +56,15 @@ private void ResetIndentLevel()
...
         }
 
+        private bool IsGeneratedTypeInternal()
+        {
+            if (Definitions == null)
+                return false;
+
+            return Definitions.IsInternal ||
+                   Definitions.Implements?.Any(it => !it.GetTypeInfo().IsVisible) == true;
+        }
```
DocSvartz, line 66: @andrerav  Have you checked if this has a negative impact on the generation of [Extension mappers? ](https://mapstermapper.github.io/Mapster/articles/tools/mapster-tool/Fluent-API-Code-generation.html#generate-using-generatemapper)

I haven't figured out what's actually going on there yet.

Origin creation logic is absolutely correct for generating class-based mappers. There is no need for implementation or inheritance.

DocSvartz (approved):

src/ExpressionTranslator/ExpressionTranslator.cs
DocSvartz: > Maybe internal class with proper implementations should be generated even in case MapperAttribute.IsInternal hasn't been specified

I think this statement contradicts the logic of C# itself. 
Implemention an internal interface doesn't force a class to be internal, since implementation is not inheritance 😅

PR #1011, opened 2026-08-20
Improvements for Mapster Tool  

### LINKED ISSUE #399: mapster tool generates non-public interface method/property implementation

DocSvartz:
Improvements for [generating mappers from interface](https://mapstermapper.github.io/Mapster/articles/tools/mapster-tool/Interface-base-Code-generation.html#generate-mapper-from-interface) :
1. Added support mappers that use internal methods or properties.
2. Added support for marking generated mapper attributes `[MapsterToolGeneratedMapper]`
	new Mapster.Tool command `mapper` option:
	- -h: true - Create a mapper with helpers attributes
	- -H: - add additional namespace for generated helpers attributes

> [!NOTE]
> By default helpers attributes generate in namespace `Mapster.Generated.Attributes.{Filename of assembly} ` specified in the `-a` option.
`-H: "MyNameSpace"` - create next namespace `Mapster.Generated.Attributes.MyNameSpace `
>If assembly filename or the string passed as a parameter ` -H` is not a valid C# namespace, a additional namespace will be created automatically.

### DIFF
--- a/src/ExpressionTranslator/ExpressionTranslator.cs
+++ b/src/ExpressionTranslator/ExpressionTranslator.cs
@@ -1268,6 +1268,81 @@ namespace ExpressionDebugger
             }
         }
 
+        public Expression VisitLambdaForGenerateMappers(LambdaExpression node, LambdaType type, Type InterfaceType, string? methodName = null,
+           bool isInternal = false)
+        {
+            VisitLambda(node, type, methodName, isInternal);
+
+            if (type == LambdaType.PrivateLambda || type == LambdaType.PublicLambda)
+            {
+                _inlineCount++;
+                if (type == LambdaType.PublicLambda)
+                {
+                    var name = methodName != null ? $"{InterfaceType.FullName}.{methodName}" : "Main";
+                    WriteLine();
+                    var funcType = MakeDelegateType(node.ReturnType, node.Parameters.Select(it => it.Type).ToArray());
+                    var exprType = typeof(Expression<>).MakeGenericType(funcType);
+                    Write(Translate(exprType), " ", name, " => ");
+                }
+
+                IList<ParameterExpression> args;
+                if (node.Parameters.Count == 1)
+                {
+                    args = new List<ParameterExpression>();
+                    var arg = VisitParameter(node.Parameters[0]);
+                    args.Add((ParameterExpression)arg);
+                }
+                else
+                {
+                    args = VisitArguments("(", node.Parameters.ToList(), p => (ParameterExpression)VisitParameter(p),
+                        ")");
+                }
+
+                Write(" => ");
+                var body = VisitGroup(node.Body, ExpressionType.Quote);
+                if (type == LambdaType.PublicLambda)
+                    Write(";");
+                _inlineCount--;
+                return Expression.Lambda(body, node.Name, node.TailCall, args);
+            }
+            else
+            {
+                var name = methodName != null ? $"{InterfaceType.FullName}.{methodName}" : "Main";
+                if (type == LambdaType.PublicMethod || type == LambdaType.ExtensionMethod)
+                {
+                    if (!isInternal)
+                        isInternal = node.ReturnType.GetTypeInfo().IsNotPublic ||
+                                     node.Parameters.Any(it => it.Type.GetTypeInfo().IsNotPublic);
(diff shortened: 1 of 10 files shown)


DocSvartz:
This solution contains a non-final part using an attribute. It can be disabled.

DocSvartz:
@andrerav 
That is, if you emulate a cases that is not currently supported:


```
 public interface IMyTypeMapper
 {
     internal AddressDTO Map(Address p1);
     internal Expression<Func<AddressDTO, Address>> Projection { get; }
 }
```

or 

```
internal interface IMyTypeMapper
{
    internal AddressDTO Map(Address p1);
    internal Expression<Func<AddressDTO, Address>> Projection { get; }
}
```

emulation step

1. replace Address to internal

`internal class Address`

+ 

2. in TypeDefinitions set 

`IsInternal = false,`

we get valid mapper  : 


```
public partial class CustomerMapper : IMyTypeMapper 
{
    internal AddressDTO Map(Address p1)
    {
        return p1 == null ? null : new AddressDTO()
        {
            Id = p1.Id,
            City = p1.City,
            Country = p1.Country
        };
    }
    AddressDTO TemplateTest.IMyTypeMapper.Map(Address p1)
    {
        return p1 == null ? null : new AddressDTO()
        {
            Id = p1.Id,
            City = p1.City,
            Country = p1.Country
        };
    }
    internal Expression<Func<AddressDTO, Address>> Projection => p2 => new Address()
    {
        Id = p2.Id,
        City = p2.City,
        Country = p2.Country
    };
    Expression<Func<AddressDTO, Address>> TemplateTest.IMyTypeMapper.Projection => p2 => new Address()
    {
        Id = p2.Id,
        City = p2.City,
        Country = p2.Country
    };
}

```

I thought this shouldn't work, but the assembly with this class compiles and the mapper itself works when cast to the interface.🤔

DocSvartz:
@andrerav  If I'm not mistaken in the current implementation😃

Now mapper members are created correctly depending on the [public/internal access modifiers specified for it or using for maping internal clases](https://github.com/DocSvartz/Mapster/blob/2a7509cdea7c3f93aaeed633456b0798f75646d4/src/TemplateTest/CreateMapExpressionTest.cs#L100).

And IsInternal continues to work as before, forcing the creation of an internal mapper with an internal members.

andrerav (approved):

andrerav:
@DocSvartz Your approach is more general and solves both #399 and more. I closed my own PR, which had issues as you pointed out.

DocSvartz:
@andrerav Perhaps it's worth adding `[MapsterToolGeneratedMapper]` to Mapster.Core instead of generating it for each dotnet mapster mapper call?

Due to the use of the` [Mapper]` attribute on the interface, it is required to use at least Mapster.Core.

DocSvartz:
> @andrerav Perhaps it's worth adding [MapsterToolGeneratedMapper] to Mapster.Core instead of generating it for each dotnet mapster mapper call?
> 
> Due to the use of the [Mapper] attribute on the interface, it is required to use at least Mapster.Core.

I'm merging this for the pre-release publish. If needed, it can be  change in the release.

PR #1018, opened 2026-09-04
Fix Issue #1017 and refactoring Mapster.Tool for ease of testing

### LINKED ISSUE #1017: Mapster.Tool - PropertySettingBuilder.Map(member, mapFunc, targetPropertyName) method is not working as expected!
I'm using - 
Mapster 10.0.12
Mapster.Tool 10.0.12
.NET 10.0.400

I have the following entity model -

```C#
public class User
{
    public int Id { get; set; }
    public string Email { get; set; }
    public string FirstName { get; set; }
    public string LastName { get; set; }
    public int Age { get; set; }
}
```
and I wanted the following DTO model -

```C#
public partial class UserDto
{
    public int Id { get; set; }
    public string Email { get; set; }
    public string FullName { get; set; }
    public int Age { get; set; }
}
```
and a mapper extension method to be generated which would map `User.FirstName + " " + User.LastName`  to `UserDto.FullName`.

I used the following code - 

```C#
public class UserCodeGenConfig : ICodeGenerationRegister
{
    public void Register(CodeGenerationConfig config)
    {
        config.AdaptTo("[name]Dto", MapType.Map)
            .ForType<User>(p =>
            {
                p.Ignore(s => s.FirstName);
                p.Map(s => s.LastName, s => $"{s.FirstName} {s.LastName}", "FullName");
            });

        config.GenerateMapper("[name]Mapper")
            .ForType<User>();
    }
}
```
and used the model and extension generating commands with Mapster.Tool. Following are the generated codes -

```C#
public partial class UserDto
{
    public int Id { get; set; }
    public string Email { get; set; }
    public string FullName { get; set; }
    public int Age { get; set; }
}
```
```C#
public static partial class UserMapper
{
    public static UserDto AdaptToDto(this User p1)
    {
        return p1 == null ? null : new UserDto()
        {
            Id = p1.Id,
            Email = p1.Email,
            FullName = p1.LastName,
            Age = p1.Age
        };
    }
}
```
As you can see, the DTO model is generated as expected, but the mapping code is exactly what it would generate if I had used -

`p.Map(s => s.LastName, "FullName");`

instead of -

`p.Map(s => s.LastName, s => $"{s.FirstName} {s.LastName}", "FullName");`

Following is the source code documentation for the `Map` method in question -

```C#
/// <summary>
/// Map a specific property of the source type to a target property using a custom mapping function.
/// </summary>
/// <typeparam name="TReturn">Type of source property.</typeparam>
/// <typeparam name="TReturn2">Type of target property type.</typeparam>
/// <param name="member">A lambda expression that identifies the source property to be mapped.</param>
/// <param name="mapFunc">A lambda expression that defines the custom mapping function.</param>
/// <param name="targetPropertyName">The name of the target property to which the source property should be mapped.</param>
/// <returns></returns>
public PropertySettingBuilder<T> Map<TReturn, TReturn2>(Expression<Func<T, TReturn>> member, Expression<Func<T, TReturn2>> mapFunc, string? targetPropertyName = null);
```

So, it seems the tool is not respecting the 2nd parameter (`Expression<Func<T, TReturn2>> mapFunc`) of this `Map` method.

Is there anything I failed to understand or is this a bug? Is there any other approach to achieve the expected result?

DocSvartz:


DocSvartz:
Reopen and rebuilding in #1019

PR #1021, opened 2026-09-18
perf: materialize hidden-member comparison names

vb-kalei:
The patch and text below are AI generated, but I submit them with the hope that they are useful. In our (big) project the compilation of our main mappers takes 4 minutes, this change reduced it to 1 minute. If you need anything else, please let me know!

### Problem

ReflectionUtils.DropHiddenMembers<T> creates a deferred intersection of member names, then calls Contains(member.Name) on that sequence for each member. Those calls repeatedly rebuild intersection state and revisit the source.
A deterministic regression test using three reflected properties reproduces the issue: full enumeration visits the source nine times, exceeding the expected bound of six.

### Solution

Materialize the existing comparison-name sequence once per enumeration into a HashSet<string> using StringComparer.Ordinal.
The remainder of the method is unchanged.
Behavior and scope
•	Preserves case-sensitive matching, output order, duplicates, and first-current-member/MetadataToken selection for stable inputs.
•	Preserves deferred execution and recomputation on subsequent enumerations.
•	Materialization changes traversal timing and can increase upfront work when enumeration stops early.
•	Does not make the entire method linear: the existing currentTypeMembers.First(...) lookup remains.
•	No reflection caching, metadata-token identity changes, public API changes, dependency changes, or unrelated refactoring.

### Tests

Adds ten MSTest/Shouldly tests covering hidden properties and fields, private-property/public-field hiding, case sensitivity, ordering and duplicates, empty inputs, deferred execution and recomputation, deterministic enumeration cost, first-current-member selection, and compiled mapping to new and existing destinations.

Reproduction:
```dotnet test src/Mapster.Tests/Mapster.Tests.csproj -c Release -f net10.0 --filter FullyQualifiedName~WhenDroppingHiddenMembers --logger "console;verbosity=normal"```

Against upstream 4a8aaa4 in an isolated worktree: nine tests pass and the enumeration-cost regression fails with nine visits versus a bound of six. With this patch: all ten pass.


vb-kalei:
@DocSvartz Please note, that I uploaded a better version of this as #1023. If you like that one, this one can be declined and closed.

DocSvartz:
Closed. because #1023 was merged.

